PR #1, opened 2022-12-08
Migrate benchmark command from as benchmark-url

felixarntz:
Originally from https://github.com/WordPress/performance/blob/3cc0edcf5d495adbcd2dfede1c4ffdda73b97c6d/bin/plugin/commands/benchmark.js, implemented by @eugene-manuilov.

This is better suited for within this repository than the Performance Lab one, as it's not strictly related to the plugin.

The name is changed from `benchmark` to `benchmark-url`, to be a bit more specific what this command is for.

This PR also updates the code to fit into this repository (e.g. ES imports/exports, using the `table` helper function which simplifies the code here a bit).

The output should 100% match what the original command would produce.

### DIFF
--- a/cli/commands/benchmark-url.mjs
+++ b/cli/commands/benchmark-url.mjs
@@ -1,0 +1,259 @@
+/**
+ * CLI command to benchmark several URLs for TTFB and Server-Timing metrics.
+ *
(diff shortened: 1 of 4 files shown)


eugene-manuilov (approved):
LGTM

PR #3, opened 2022-12-09
Analysis of hosting providers influence on the response time.

eugene-manuilov:
This PR contains some changes to the benchmark command that were needed to perform the analysis of hosting providers. ~~It also includes a GitHub action that actually runs analysis for provided URLs and saves results as artifacts.~~


cli/commands/benchmark-url.mjs
```diff
@@ -245,6 +247,10 @@ function outputResults( opt, results ) {
 			tableRow.push( metricAvgMs );
 		}
 
+		for ( let i = tableRow.length; i < headings.length; i++ ) {
+			tableRow.push( '' );
+		}
```
felixarntz, line 252: Hmm, this doesn't look right. In what situation would there be missing columns? _If_ there are, we would need to match them with the right index based on the heading, not just fill up empty columns at the end, since otherwise the metric values would be associated with the wrong metric heading.

eugene-manuilov: @felixarntz in my analysis I have a mix of URLs, part of them lead to empty PHP scripts, another part to WordPress pages. Responses for simple PHP scripts don't contain Server-Timing headers, thus I need to add empty strings to make sure I have equal number or rows for all URLs.

felixarntz:
@eugene-manuilov Do you still plan on completing this PR?

cli/commands/benchmark-url.mjs
eugene-manuilov: You are right. Updated.

eugene-manuilov:
> @eugene-manuilov Do you still plan on completing this PR?

@felixarntz, yes, sorry for the delay. PR is updated. Please, take a look again.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,26 @@
...
+          - shared-1
+          - shared-2
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          cache: npm
+      - run: npm ci
+      - run: npm run benchmark --silent -- -c 2 -n 10 -f analysis/hosting-stack/${{ matrix.file }}.txt -o csv > ${{ matrix.file }}.csv
```
felixarntz, line 21: Instead of using specific files here, would it be possible to provide the files somehow when dispatching the workflow?

If possible, I would prefer not to include the files with the concrete URLs in the repo. Would be great if those could be provided on the fly, not in source code.

analysis/hosting-stack/managed-1.txt
```diff
@@ -0,0 +1,20 @@
+https://site1-siteground.cmsdevrel.com/wp-content/index.php
```
felixarntz, line 1: See above, if at all possible, it would be great to remove all those files with the concrete lists of URLs.

package.json
```diff
@@ -15,6 +15,7 @@
   "scripts": {
     "format": "wp-scripts format ./cli/**/*.mjs",
     "lint": "wp-scripts lint-js ./cli --ext mjs",
-    "research": "./cli/run.mjs"
+    "research": "./cli/run.mjs",
+    "benchmark": "npm run research --silent -- benchmark-url"
```
felixarntz, line 19: Please remove this for now, other commands don't have this shorthand either. What would be great instead though is to document the benchmark command in `cli/README.md`, similar to how the other commands are documented there.

felixarntz (commented):
@eugene-manuilov The benchmark command itself looks great now, left a bit of other feedback below.

.github/workflows/benchmark.yml
eugene-manuilov: Yes, I agree with this, but unfortunately this is not possible. Probably we should remove this workflow and URL files, right?

analysis/hosting-stack/managed-1.txt
eugene-manuilov: Replied in the above comment.

package.json
eugene-manuilov: Removed and updated the readme file.

.github/workflows/benchmark.yml
felixarntz: @eugene-manuilov Agreed, in that case probably this PR should only have the benchmark command changes.

Can you move this workflow file (not the URL lists) into a public GitHub Gist on your account maybe? If you do that, I think it would be great to keep a reference to that in here, for example you could add a code comment somewhere in the `benchmark-url.js` file that links to this Gist (e.g. "Example for how to use this command in a GitHub workflow: {url}").

cli/README.md
```diff
@@ -94,3 +94,27 @@ Get Server-Timing header medians _and_ all individual run values:
 ```
 wpt-server-timing --test 221011_AiDcV7_GGM --include-runs
 ```
+
+### `benchmark-url`
+
+Sends the selected number of requests with a certain concurrency to provided URLs to find out the median response time for each URL. It can also track Server-Timing metrics and get media values for each of them.
```
felixarntz, line 100:
```suggestion
Sends the selected number of requests with a certain concurrency to provided URLs to find out the median response time for each URL. It can also track Server-Timing metrics and get median values for each of them.
```

cli/README.md
```diff
@@ -94,3 +94,27 @@ Get Server-Timing header medians _and_ all individual run values:
...
+
+#### Arguments
+
+* `--url` (`-u`): An URL to benchmark.
+* `--concurrency` (`-c`): Number of requests to make at the same time.
+* `--number` (`-n`): Total number of requests to send.
+* `--file` (`-f`): File with URLs to run benchmark tests for.
```
felixarntz, line 107: Maybe add an explanation on the required file format? (simply one URL per line)

cli/README.md
```diff
@@ -94,3 +94,27 @@ Get Server-Timing header medians _and_ all individual run values:
...
+
+Send 10 request, 2 requests at the same time:
+```
+benchmark-url --url https://example.com/ -n 10 -c 2
+```
+
+Same as above, but results are formatted as CSV:
+```
+benchmark-url --url https://example.com/ -n 10 -c 2 --output csv
+```
```
felixarntz, line 120: Maybe add one more example for using the `--file` argument? (Can refer to whatever non-existing file name)

cli/commands/benchmark-url.mjs
```diff
@@ -210,44 +220,37 @@ function outputResults( opt, results ) {
...
-		tableData.push( tableRow );
+		tableData.push( [
+			url,
+			`${ completionRate }%`,
+			round( calcMedian( responseTimes ), 2 ),
+			...Object.values( vals ),
+		] );
 	}
 
+	console.log( headings	,tableData );
```
felixarntz, line 252: I think this is from debugging and should be removed?

cli/commands/benchmark-url.mjs
```diff
@@ -133,6 +133,7 @@ function benchmarkURL( params ) {
 	const metrics = {};
```
felixarntz, line 133: See my previous reply, maybe add a comment somewhere on top of this file linking to a Gist with the GitHub workflow?

felixarntz (commented):
@eugene-manuilov A few last comments here.

Once you've removed the changes around the GH workflow and URL list files, this should be good to go.

.github/workflows/benchmark.yml
eugene-manuilov: Ok, removed.

cli/README.md
eugene-manuilov: Yes, added.

cli/README.md
eugene-manuilov: Yep, added.

cli/commands/benchmark-url.mjs
eugene-manuilov: Yeah, this is for debugging, removed 😄

cli/commands/benchmark-url.mjs
eugene-manuilov: Yep, sounds good to me. Added.

felixarntz (approved):
@eugene-manuilov Thanks, LGTM!

felixarntz:
@mukeshpanchal27 Can you please give this another review (and approval eventually)?

mukeshpanchal27 (approved):

PR #24, opened 2023-01-05
Add query: Queries to assess usage of defer on scripts

ankitrox:
*  `sql/2023/01/percentage-of-deferred-scripts.sql` will process 2.96 TB of data, so did not run to add the results here.

### DIFF
--- a/sql/2023/01/percentage-of-deferred-scripts.sql
+++ b/sql/2023/01/percentage-of-deferred-scripts.sql
@@ -1,0 +1,43 @@
+# HTTP Archive query to get % of WordPress sites that use defer on any script.
+#
+# WPP Research, Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+SELECT
+  har_pages._TABLE_SUFFIX AS client,
+  COUNTIF(CAST(JSON_EXTRACT(JSON_EXTRACT_SCALAR(payload, '$._javascript'), '$.script_tags.async') AS INT64) > 0) AS async,
+  COUNTIF(CAST(JSON_EXTRACT(JSON_EXTRACT_SCALAR(payload, '$._javascript'), '$.script_tags.defer') AS INT64) > 0) AS defer,
+  COUNTIF(CAST(JSON_EXTRACT(JSON_EXTRACT_SCALAR(payload, '$._javascript'), '$.script_tags.async_and_defer') AS INT64) > 0) AS async_and_defer,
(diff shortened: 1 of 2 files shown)


sql/2023/01/atleast-one-deferred-script.sql
```diff
@@ -0,0 +1,41 @@
...
+  GROUP BY
+    client,
+    page),
+  UNNEST([10, 25, 50, 75, 90, 100]) AS percentile
+GROUP BY
+  percentile,
+  client
+ORDER BY
+  client,
+  percentile
```
felixarntz, line 41: This query is a bit unnecessarily complex. For example, it includes a distribution, but we don't need that here. Also the `getDeferredScripts` temporary function is unnecessary since we can already do that in one line via `JSON_EXTRACT`.

https://github.com/GoogleChromeLabs/wpp-research/pull/29/files#diff-d647443f186ed55d494b9389bc8322970137af7a407a69bcc8109c370cff0001 includes the metric we're looking for and is much simpler. So I think we can remove this query from the PR.

sql/2023/01/no-scripts-on-page.sql
```diff
@@ -0,0 +1,33 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SELECT
+  har_pages._TABLE_SUFFIX AS client,
+  COUNTIF(CAST(JSON_EXTRACT(JSON_EXTRACT_SCALAR(payload, '$._javascript'), '$.script_tags.total') AS INT64) = 0) / COUNT(0) AS total_pct,
```
felixarntz, line 19: This query gets the number of sites with no `script` tags at all on the page, which may be useful in theory. That includes inline scripts. But since we are researching the usage of `defer`, and inline scripts cannot be deferred anyway, I think we are looking for sites with no _external_ `script` tags actually.

Similarly to the above, https://github.com/GoogleChromeLabs/wpp-research/pull/29/files#diff-d647443f186ed55d494b9389bc8322970137af7a407a69bcc8109c370cff0001 already includes the number of sites with _any_ external scripts, so that means we already have the metric we need (`total_wp_sites - with_any_external_scripts = without_any_external_scripts`). So I think we can remove this query as well.

sql/2023/01/percentage-of-deferred-scripts.sql
```diff
@@ -0,0 +1,32 @@
...
+    ON
+        har_tech.url = har_pages.url
+    WHERE
+        har_pages._TABLE_SUFFIX = har_tech._TABLE_SUFFIX
+    AND
+        app = 'WordPress'
+    AND
+        category = 'CMS'
+GROUP BY
+  client
```
felixarntz, line 32: This query in particular is what we wanted to have a distribution for, however here it only gets a single percentage value. https://github.com/GoogleChromeLabs/wpp-research/pull/29/files#diff-4e1721f7172199df4fe45a80b994617ae36fea1d4841115d3e0881fa2262047d provides this distribution.

felixarntz (commented):
@ankitrox It's great that you came up with the queries here, some of them are close to what we are looking for.

However, it looks like #29 includes all the metrics we need in a more efficient way; so I think we should close this PR.

Please review #29 though, and if you have any recommendations to improve things there or incorporate something from here in that PR let me know!

felixarntz:
@ankitrox Can you please review my feedback here?

felixarntz:
Closing in favor of the merged #29.

PR #15, opened 2022-12-16
Add Queries to access usage of fetchpriority='high' on LCP image

kt-12:
Fixes #5 

> **Note:** *At the time of running this query, the latest httparchive reports available were of `01, November 2022`, which were used in all the queries here*.
----
 

> **Overall Opportunity Score %** = *100 x ( No. of WordPress sites having LCP image  **without** fetchpriority='high') / (**Total no. of** WordPress sites)*
### Result
client | with_fetchpriority_on_lcp | without_fetchpriority_on_lcp | total_with_lcp | total_wp_sites | opportunity | overall_opportunity
-- | -- | -- | -- | -- | -- | --
desktop | 444 | 1818969 | 1819413 | 4245886 | 99.976 % | 42.841 %
mobile | 628 | 2189295 | 2189923 | 5707622 | 99.971 % | 38.357 %

PR #27, opened 2023-01-07
Introduce alternative, slightly more efficient version of fetchpriority opportunity query

felixarntz:
This query is a slightly more efficient version of the query introduced in #15. The results are pretty much the same, except for some extremely small differences that are natural given that one uses the overall table while this one uses the monthly tables.

One thing to note is that for now this query is set for `October 1, 2022` results instead of `November 1, 2022`, since for some reason HTTP Archive is currently still missing the mobile results for November. That shouldn't matter too much though, we can always rerun the query later for a newer dataset.

This version of the query is heavily inspired by https://discuss.httparchive.org/t/lazy-loaded-lcp-images-by-wordpress-version/2455/2.

Comparing the two queries in terms of efficiency (when running both queries against the October dataset), both take roughly the same time to complete. However this query consumes only 4.18 TB while the other one consumes 7.91 TB.

This PR does not overwrite the original query as it is worth keeping both versions for reference.

## Query results

Row | client | with_fetchpriority_on_lcp | without_fetchpriority_on_lcp | total_with_lcp | total_wp_sites | pct_opportunity | pct_overall_opportunity
-- | -- | -- | -- | -- | -- | -- | --
1 | desktop | 284 | 1503490 | 1503774 | 3502837 | 0.99981114183381281 | 0.42922065742710835
2 | mobile | 337 | 2091778 | 2092115 | 5465547 | 0.99983891898867894 | 0.382720704

For comparison, the results for the query from #15 for the same month are:

Row | client | with_fetchpriority_on_lcp | without_fetchpriority_on_lcp | total_with_lcp | total_wp_sites | opportunity | overall_opportunity
-- | -- | -- | -- | -- | -- | -- | --
1 | desktop | 284 | 1504820 | 1505104 | 3505796 | 99.981 % | 42.924 %
2 | mobile | 337 | 2092862 | 2093199 | 5468298 | 99.984 % | 38.273 %

Based on October 2022 dataset

### DIFF
--- a/sql/2023/01/lcp-image-without-fetchpriority-high-opportunity-more-efficient.sql
+++ b/sql/2023/01/lcp-image-without-fetchpriority-high-opportunity-more-efficient.sql
@@ -45,10 +45,10 @@ WITH
 SELECT
   client,
   COUNTIF(fetchpriority = "high"
-    AND nodeName = "IMG" ) AS `with_fetchpriority_on_lcp`,
+    AND nodeName = "IMG" ) AS with_fetchpriority_on_lcp,
   COUNTIF(nodeName = "IMG") - COUNTIF(fetchpriority = "high"
-    AND nodeName = "IMG" ) AS `without_fetchpriority_on_lcp`,
-  COUNTIF(nodeName = "IMG") AS `total_with_lcp`,
+    AND nodeName = "IMG" ) AS without_fetchpriority_on_lcp,
+  COUNTIF(nodeName = "IMG") AS total_with_lcp,
   COUNT(0) AS total_wp_sites,
   (COUNTIF(nodeName = "IMG") - COUNTIF(fetchpriority = "high"
     AND nodeName = "IMG" )) / COUNTIF(nodeName = "IMG") AS pct_opportunity,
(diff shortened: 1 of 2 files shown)


felixarntz:
@kt-12 Reviewed https://discuss.httparchive.org/t/lazy-loaded-lcp-images-by-wordpress-version/2455 today and tried to use that approach for an alternative version of your fetchpriority query. Here's the result.

TLDR it is slightly more efficient, results are almost the same. I think we should keep both for reference purposes, it shows two different ways to achieve a similar goal.

kt-12:
> @kt-12 Reviewed https://discuss.httparchive.org/t/lazy-loaded-lcp-images-by-wordpress-version/2455 today and tried to use that approach for an alternative version of your fetchpriority query. Here's the result.
> 
> TLDR it is slightly more efficient, results are almost the same. I think we should keep both for reference purposes, it shows two different ways to achieve a similar goal.

Thanks @felixarntz. I happened to review the exact same query today for another task. It makes more sense as we don't have to traverse a very large table or do join with the `technologies` table just to know if it's a WordPress site. 

The same logic can be extended while working with the Lighthouse report. Lighthouse report has "stackPacks" array and the very first object if it was a WordPress site is of WordPress. We can use this to avoid joining with another table
```JSON
    "stackPacks": [
        {
            "id": "wordpress",
            "title": "WordPress",
```

kt-12 (dismissed):
Thanks @felixarntz , LGTM!

kt-12 (approved):

kt-12 (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

PR #28, opened 2023-01-07
Implement query to assess lazy-loaded LCP image fix opportunity

### LINKED ISSUE #8: Query to assess lazy-loading of LCP image
* Create HTTP Archive query to get **% of WordPress sites that use `loading="lazy"` on LCP image**
    * Relate it with query from #5 for **% of WordPress sites that even have an image as the LCP element**
* Create HTTP Archive query to get **% of WordPress sites that are on version >= 5.5 _and_ have at least one image with `loading="lazy"`**

felixarntz:
Fixes #8 

Inspired by https://discuss.httparchive.org/t/lazy-loaded-lcp-images-by-wordpress-version/2455/2, this query allows to assess the opportunity of fixing the remaining problems where WordPress sites lazy-load their LCP image.

The query includes the opportunity relative to the number of WordPress sites that actually have an LCP image, but also relative to the total number of WordPress sites. For reference, the query also shows the number of sites on a WordPress version greater than or equal to 5.5, since that is when lazy-loading was introduced in WordPress core.

## Query results

Row | client | with_lazyloaded_lcp_image | with_lcp_image | total_wp_sites | pct_lcp_image_opportunity | pct_total_opportunity | wp_gte_55
-- | -- | -- | -- | -- | -- | -- | --
1 | desktop | 321428 | 1503774 | 3502837 | 0.21374754451134279 | 0.091762191617822925 | 3164718
2 | mobile | 434920 | 2092115 | 5465547 | 0.20788532179158412 | 0.079574834870142 | 4857803

Based on October 2022 dataset

### DIFF
--- a/sql/2023/01/lazyloaded-lcp-opportunity.sql
+++ b/sql/2023/01/lazyloaded-lcp-opportunity.sql
@@ -1,0 +1,66 @@
+# HTTP Archive query to get % of WordPress sites that lazy-load their LCP image.
+#
+# WPP Research, Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/28
+CREATE TEMP FUNCTION
+  getLoadingAttr(attributes STRING)
+  RETURNS STRING
+  LANGUAGE js AS '''
+  try {
+    const data = JSON.parse(attributes);
+    const loadingAttr = data.find(attr => attr["name"] === "loading")
+    return loadingAttr.value
+  } catch (e) {
+    return "";
+  }
+''';
+
+WITH
+  lcp_stats AS (
+  SELECT
+    _TABLE_SUFFIX AS client,
+    url,
+    JSON_EXTRACT_SCALAR(payload, '$._performance.lcp_elem_stats.nodeName') AS nodeName,
+    JSON_EXTRACT_SCALAR(payload, '$._performance.lcp_elem_stats.url') AS elementUrl,
+    JSON_EXTRACT(payload, '$._performance.lcp_elem_stats.attributes') AS attributes,
+    JSON_EXTRACT(payload, '$._detected_apps.WordPress') AS wpVersion,
+    getLoadingAttr(JSON_EXTRACT(payload, '$._performance.lcp_elem_stats.attributes')) AS loading,
+  FROM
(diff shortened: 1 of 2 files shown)


vishalkakadiya:
@felixarntz The query is more significant than 1 TB of data processing, so I haven't run it, but it looks good. Thank you!

vishalkakadiya (approved):

sql/2023/01/lazyloaded-lcp-opportunity.sql
```diff
@@ -0,0 +1,66 @@
...
+  SELECT
+    _TABLE_SUFFIX AS client,
+    url,
+    JSON_EXTRACT_SCALAR(payload, '$._performance.lcp_elem_stats.nodeName') AS nodeName,
+    JSON_EXTRACT_SCALAR(payload, '$._performance.lcp_elem_stats.url') AS elementUrl,
+    JSON_EXTRACT(payload, '$._performance.lcp_elem_stats.attributes') AS attributes,
+    JSON_EXTRACT(payload, '$._detected_apps.WordPress') AS wpVersion,
+    getLoadingAttr(JSON_EXTRACT(payload, '$._performance.lcp_elem_stats.attributes')) AS loading,
+  FROM
+    `httparchive.pages.2022_10_01_*`
```
mukeshpanchal27, line 42:
```suggestion
    `httparchive.pages.2022_11_01_*`
```

Should use November 2022 data as it now ready for use.

sql/2023/01/lazyloaded-lcp-opportunity.sql
```diff
@@ -0,0 +1,66 @@
...
+    AND nodeName = "IMG") AS with_lazyloaded_lcp_image,
+  COUNTIF(nodeName = "IMG") AS with_lcp_image,
+  COUNT(0) AS total_wp_sites,
+  COUNTIF(loading = "lazy"
+    AND nodeName = "IMG") / COUNTIF(nodeName = "IMG") AS pct_lcp_image_opportunity,
+  COUNTIF(loading = "lazy"
+    AND nodeName = "IMG") / COUNT(0) AS pct_total_opportunity,
+  # For reference, include number of sites greater than or equal to WP 5.5. Sites without a known version can be considered
+  # part of this since at this point they are most likely on a more recent version.
+  COUNTIF(wpVersion = '""'
```
mukeshpanchal27, line 57: Because the query's where clause only returns results that aren't null, does `wpVersion = '""'` work?

mukeshpanchal27 (commented):
Thanks @felixarntz, Left some feedback.

kt-12 (approved):
LGTM!, Thank you, @felixarntz  and @vishalkakadiya.

sql/2023/01/lazyloaded-lcp-opportunity.sql
felixarntz: @mukeshpanchal27 Yes, this is looking for an empty string as the version, which is different from `null`.

sql/2023/01/lazyloaded-lcp-opportunity.sql
felixarntz: @mukeshpanchal27 Fair point, but we don't necessarily have to do that. We are using the October data already in the analysis, so changing it wouldn't be good at this point.

FWIW all of the queries in this repo will have some specific month on them. That's okay, we shouldn't keep this up to date. Just using what's the latest when we write the query is a good approach.

PR #40, opened 2023-02-17
Add command to benchmark Web Vitals via `web-vitals` library

felixarntz:
This introduces a new `benchmark-web-vitals` command that, similar to the existing `benchmark-url` command allows making 1 or more requests to 1 or more URLs, capture the results and compute the medians for each metric. However, the notable difference is that this one measures Core Web Vitals and other key client metrics, using https://github.com/GoogleChrome/web-vitals.

As part of this PR, the `benchmark-url` command is renamed to a more appropriate `benchmark-server-timing`, to clearly differentiate the two commands and clarify the different purposes. They also function quite differently, as the existing command simply makes a `curl` request to the URL while the new one here uses a headless browser.

The initial iteration implemented in this PR only includes load time metrics (FCP, LCP, and TTFB) as those are the most critical for us to measure, and it appears that they are easier to implement. An early draft PR for adding other Web Vitals CLS, FID, and INP exists at #41 and can be explored afterwards as an iteration.

### Testing

You can test this command by following the instructions included in this PR (see https://github.com/GoogleChromeLabs/wpp-research/tree/add/benchmark-web-vitals/cli#benchmark-web-vitals), which also provide examples.


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -0,0 +1,216 @@
...
+	}
+}
+
+async function benchmarkURL( browser, params ) {
+	/*
+	 * For now this only includes load time metrics.
+	 * In the future, additional Web Vitals like CLS, FID, and INP should be
+	 * added, however they are slightly more complex to retrieve through an
+	 * automated headless browser test.
+	 * See https://github.com/GoogleChromeLabs/wpp-research/pull/41.
```
tunetheweb, line 105: FID and INP are more complex as require an interaction, but the load CLS should be similar to FCP/LCP/TTFB shouldn't it?

Granted, that often isn't the full CLS of a page (as it can be impacted by scroll, or other interactions), but it's still usually a good part of CLS. And that "load CLS" is what Lighthouse and PSI already captures despite it's limitations.

The only thing is that, kind of like LCP, CLS is not emitted by web-vitals.js, until the page is hidden (including navigating away from it). For the web-vitals.js library we dummy this visibility change:
https://github.com/GoogleChrome/web-vitals/blob/main/test/utils/stubVisibilityChange.js

And then call it like this:
```js
    // Wait until all images are loaded and rendered, then change to hidden.
    await imagesPainted();
    await stubVisibilityChange('hidden');
```

Maybe that's enough to make it too complicated here?

The alternative is to use the `reportAllChanges` flag and only report at the end of the test.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -0,0 +1,216 @@
...
+
+		completeRequests++;
+
+		await Promise.all(
+			Object.values( metricsDefinition ).map( async ( value ) => {
+				// Wait until global is populated.
+				await page.waitForFunction( `window.${ value.global } !== undefined` );
+
+				// Do a random click, since only that seems to trigger certain metrics like LCP.
+				await page.click( 'body' );
```
tunetheweb, line 159: FYI, this is [by design](https://web.dev/lcp/#:~:text=The%20browser%20will%20stop%20reporting%20new%20entries%20as%20soon%20as%20the%20user%20interacts%20with%20the%20page%20(via%20a%20tap%2C%20scroll%2C%20or%20keypress)%2C%20as%20user%20interaction%20often%20changes%20what%27s%20visible%20to%20the%20user%20(which%20is%20especially%20true%20with%20scrolling).)

As mentioned above, CLS is also only reported on visibility change.
FID requires an interaction.
INP requires an interaction AND a visibility change.

cli/commands/benchmark-web-vitals.mjs
tunetheweb: Yeah sounds like it’s due to not changing the visibility so CLS not being “finalised”. Which I realised and edited after writing that first sentence (but before submitting the review).

felixarntz:
@oandregal Would you mind giving this PR a review? Have you been able to test the command yet?

vishalkakadiya (approved):
I tested the command with all the options, and it is working great. Approving PR now!

10upsimon (dismissed):
I've tested all commands as listed in the branches README, and all are functioning correctly for me. Approved from my end.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -0,0 +1,220 @@
...
+	table,
+	isValidTableFormat,
+	OUTPUT_FORMAT_TABLE,
+} from '../lib/cli/logger.mjs';
+import { calcMedian } from '../lib/util/math.mjs';
+
+export const options = [
+	{
+		argname: '-u, --url <url>',
+		description: 'An URL to run benchmark tests for',
```
10upsimon, line 41:
```suggestion
		description: 'A URL to run benchmark tests for',
```

@felixarntz I've already approved, and this is not a deal breaker, but `A` would be more grammatically correct than `An` in this instance. More of a nitpick than anything else.

felixarntz: @10upsimon Thanks, fixed! A similar mistake was present in the `benchmark-server-timing.mjs` file too so I fixed it in both places.

10upsimon (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz, Quick review and approve for merge.

vishalkakadiya (approved):

felixarntz:
Thanks @oandregal for the feedback!

Regarding LCP - TTFB, that's a great idea to add as a metric, I'll take a note to do that (though I'm AFK for the next few days, so it'll take a couple of days). Feel free to open a PR if you have capacity, but otherwise I'll get to it once I'm back.

For the variability, are those ms differences you're sharing between different runs of the same command? For example, one time you ran it with 10 runs, you got p25 set to some value, and the other time the p25 was 15ms different? I'm definitely interested in exploring this further to see how we can get those numbers down, but partially of course it will be impossible due to varying machine resources etc. For example I'd be curious how the results differed for the other percentiles in those specific runs you did. Maybe we need to conduct a more detailed analysis on how the results differ between different runs with the tool. Would be great if we could use that to find the "sweet spot" for how many runs to do per each invocation of the command.

PR #41, opened 2023-02-17
Enhance `benchmark-web-vitals` command with additional metrics

felixarntz:
Follow up to #40: The initial version of the `benchmark-web-vitals` command only comes with load time metrics, since other metrics are a bit more complicated to capture in a headless browser request, e.g. CLS and INP require interactions to happen and capture events throughout the page lifecycle, not just load time related.

**This is an early draft intended for iteration in the future.**

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -97,17 +97,29 @@ export async function handler( opt ) {
 }
 
 async function benchmarkURL( browser, params ) {
-	/*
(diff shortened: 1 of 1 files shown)


felixarntz:
Pointing to feedback from #40 that will be relevant for this:
* https://github.com/GoogleChromeLabs/wpp-research/pull/40#discussion_r1118891594
* https://github.com/GoogleChromeLabs/wpp-research/pull/40#discussion_r1118911471

PR #42, opened 2023-02-22
Add WP Profiler mu-plugin

joemcgill:
This adds an example mu-plugin that can be used for configuring a WordPress environment set up using the `wp-env` package with XHProf support ([related PR](https://github.com/WordPress/gutenberg/pull/48147)).

### DIFF
--- a/plugins/wp-profiler/.wp-env.json
+++ b/plugins/wp-profiler/.wp-env.json
@@ -1,0 +1,11 @@
+{
+    "mappings": {
+        "wp-content/mu-plugins": "./"
+    },
+    "config": {
+        "WP_DEBUG": false,
+        "WP_DEBUG_LOG": false,
+        "WP_DEBUG_DISPLAY": false,
+        "SCRIPT_DEBUG": false
+    }
+}
--- a/plugins/wp-profiler/plugin.php
+++ b/plugins/wp-profiler/plugin.php
@@ -1,0 +1,34 @@
+<?php
+require __DIR__ . '/vendor/autoload.php';
+
+use Xhgui\Profiler\Profiler;
+use Xhgui\Profiler\ProfilingFlags;
+
+// Add this block inside some bootstrapper or other "early central point in execution"
+try {
+    /**
+     * The constructor will throw an exception if the environment
+     * isn't fit for profiling (extensions missing, other problems)
+     */
+    $profiler = new Profiler(
+        array(
+            'save.handler.upload' => array(
+                // Use docker's internal networking to connect containers.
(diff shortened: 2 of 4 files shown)


plugins/wp-profiler/.wp-env.json
```diff
@@ -0,0 +1,11 @@
...
+    "mappings": {
+        "wp-content/mu-plugins": "./"
+    },
+    "config": {
+        "WP_DEBUG": false,
+        "WP_DEBUG_LOG": false,
+        "WP_DEBUG_DISPLAY": false,
+        "SCRIPT_DEBUG": false
+    }
+}
```
felixarntz, line 11: Having this file here seems a bit out of scope to me as it's for the environment, and the plugin folder itself does not make that environment work.

I think it would be great to implement a simple wp-env configuration in this repository for XHProf so that it's standalone. Of course for XHProf that requires us to first merge your PR https://github.com/WordPress/gutenberg/pull/48147.

My suggestion would be to, instead of a general `plugins` folder and this plugin, add a folder like `wp-profiling-env` which includes a basic `package.json` with the necessary dependencies and scripts to run `wp-env`, and then an `mu-plugins` folder with the `wp-profiler` plugin you have here, as that is most useful as part of that environment.

WDYT? We could of course split that work apart, since right now `@wordpress/env` does not yet support the necessary XHProf work, but I think eventually this would be a better outcome, have a standalone solution in here rather than only the MU plugin and require the rest to be set up manually.

plugins/wp-profiler/plugin.php
```diff
@@ -0,0 +1,34 @@
+<?php
+require __DIR__ . '/vendor/autoload.php';
```
felixarntz, line 2: Nit-pick? Should we have a simple plugin header here?

plugins/wp-profiler/plugin.php
```diff
@@ -0,0 +1,34 @@
...
+        )
+    );
+
+    // The profiler itself checks whether it should be enabled
+    // for request (executes lambda function from config)
+    $profiler->start();
+} catch (Exception $e){
+    // throw away or log error about profiling instantiation failure
+    error_log($e->getMessage());
+}
```
felixarntz, line 34: Nit-pick: This file doesn't follow WP coding standards (mostly indentation and spacing-wise).

felixarntz (commented):
@joemcgill Thank you, I left some feedback. Overall it looks good, though I wonder whether we could work towards having a profiling environment in this repository (not necessarily all in this one PR) rather than only the `wp-profiler` plugin.

PR #43, opened 2023-02-28
Add support for new `--show-percentiles` arg to get more granular percentiles instead of just median to all commands

felixarntz:
Follow up to #40.

This PR chooses a hard-coded set of percentiles 10, 25, 50 (median), 75, 90, previously discussed with @oandregal. This is also what is commonly used for HTTP Archive queries by that team, and we have also used that set of percentiles in our own HTTP Archive queries (e.g. [this one](https://github.com/GoogleChromeLabs/wpp-research/blob/8f7763f964ae4a05a78e9b84877f66877f6e6410/sql/2023/01/external-deferred-scripts-distribution.sql#L25)).

In the future we can maybe make the concrete percentiles controllable by the argument, but that doesn't seem necessary at this point. The above percentiles are a good baseline to get a more representative impression of the overall accuracy of the values.

### Testing this PR

1. Run each of the 4 commands prior to this change (for any URL / any WebPageTest result).
2. Switch to this PR branch.
3. Run the same commands as above again.
    * For the `wpt-*` commands, make sure the median results are the same.
    * For the `benchmark-*` commands, make sure the median results are roughly similar (they won't be the same because it's different runs).
4. Run the same commands again, but now include the `--show-percentiles` flag in each one.
    * For the `wpt-*` commands, make sure the `p50` results are the same as the previous median results.
    * For the `benchmark-*` commands, make sure the `p50` results are roughly similar to the previous median results (they won't be the same because it's different runs).
    * For any command, do a quick sanity check on whether percentile values look right, e.g. if you use 8 runs for your command, the `p25` value should be the value of the 2nd lowest run result, or if you use 10 runs, the `p90` value should be the 9th lowest / 2nd highest run result. This sanity check can be done for any of the commands, since the underlying percentile calculation logic is the same in all of them.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -24,10 +24,14 @@ This command parses the median values for given metrics out of the WebPageTest r
 
 By default, only the median values are returned. You can optionally request all the individual run values as well.
 
-#### Required arguments
+#### Arguments
 
 * `--test` (`-t`): You need to pass a WebPageTest result ID (e.g. "221011_AiDcV7_GGM") or URL (e.g. "https://www.webpagetest.org/result/221011_AiDcV7_GGM/"). You can optionally pass multiple test result IDs to merge their metrics. This is usually not relevant but can be helpful to combine multiple results with similar test configuration, to effectively have more test runs than the limit of 9 that WebPageTest imposes.
 * `--metrics` (`-m`): You need to pass one or more WebPageTest metrics. Any metrics available on the "Graph Page Data" view (e.g. "https://www.webpagetest.org/graph_page_data.php?tests=221011_AiDcV7_GGM&median_value=1") are available. For a full list, please see the source code of the `createGetSingleMetricValue_()` function in the `lib/wpt/result.mjs` file. Additionally, you can access any Server-Timing metric by its identifier prefixed with "Server-Timing:". You can even aggregate multiple metrics in one via addition (` + `) and/or subtraction (` - `). Make sure to include a space before and after the arithmetic operator.
+* `--format` (`-f`): The output format: Either "table" or "csv".
+* `--show-percentiles` (`-p`): Whether to show more granular percentiles instead of only the median.
+* `--include-runs` (`-i`): Whether to also show the full results for all runs.
+* `--rows-as-columns` (`-r`): Whether to inverse rows and columns.
 
 #### Examples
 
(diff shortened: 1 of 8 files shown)


cli/README.md
```diff
@@ -24,10 +24,14 @@ This command parses the median values for given metrics out of the WebPageTest r
 
 By default, only the median values are returned. You can optionally request all the individual run values as well.
 
-#### Required arguments
+#### Arguments
```
felixarntz, line 27: Adjusting this simply because not all arguments are required.

cli/README.md
```diff
@@ -105,7 +123,8 @@ Sends the selected number of requests with a certain concurrency to provided URL
 * `--concurrency` (`-c`): Number of requests to make at the same time.
 * `--number` (`-n`): Total number of requests to send.
 * `--file` (`-f`): File with URLs (one URL per line) to run benchmark tests for.
-* `--output` (`-o`): The output format.
+* `--output` (`-o`): The output format: Either "table" or "csv".
```
10upsimon, line 126: @felixarntz under all other "Arguments" headings I've noticed the usage of `--format` (`-f`) for the output format, but here it's defined as `--output` (`-o`). 

Is this correct? It may be correct in code (and function as `--output`) but should we be interchanging flags?

cli/README.md
```diff
@@ -133,7 +157,8 @@ Loads the provided URLs in a headless browser several times to measure median We
 * `--url` (`-u`): A URL to benchmark.
 * `--number` (`-n`): Total number of requests to send.
 * `--file` (`-f`): File with URLs (one URL per line) to run benchmark tests for.
-* `--output` (`-o`): The output format.
+* `--output` (`-o`): The output format: Either "table" or "csv".
```
10upsimon, line 160: @felixarntz same comment here WRT `--output` as opposed to `--format`

10upsimon:
@felixarntz code is good for me, and I've functionally tested and all of the commands and percentile/non percentile applications work as expected.

So really just looking for feedback on whether we need to alter `--output` to be inline with the `--format` usage as it is for the `wpt-metrics` and `wpt-server-timing` commands.

This applies to the `benchmark-server-timing` and `benchmark-web-vitals` commands.

Aside from that, good to go from my end thank you!

cli/README.md
felixarntz: @10upsimon I'm aware of this, this happened due to different people working on the commands and names diverging as such :)

I agree we should address this, but thinking better in a separate PR. Please feel free to open one!

cli/README.md
felixarntz: @10upsimon See my reply above.

10upsimon (approved):

felixarntz:
@mukeshpanchal27 Could you give this a review some time this week?

mukeshpanchal27 (approved):

PR #44, opened 2023-03-01
Fix format script to miss files, fix lint errors, and introduce GH workflow to lint

felixarntz:
The `npm run format` command was only acting on `.mjs` files two levels deep within the `cli` directory, thus excluding the `.mjs` files in other nesting levels in that directory. It was assumed that the `**` wildcard would cover any nesting level but apparently it doesn't for that command.

This PR fixes the `npm run format` script to cover all the `.mjs` files, and it also fixes related JS lint errors that are currently present.

Last but not least it adds a GitHub workflow to avoid such lint errors from getting into `main` in the future.


mukeshpanchal27:
@felixarntz https://github.com/GoogleChromeLabs/wpp-research/pull/44/commits/d2cb5a7a4e509d614b213da1aff59adb7ae7f465 fix empty line notice for GH.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,50 @@
...
+      - main
+      - 'feature/**'
+    paths:
+      - '.github/workflows/lint.yml'
+      - '**.js'
+      - '**.json'
+      - '**.mjs'
+      - '.eslint*'
+      - '.nvmrc'
+      - '**/package.json'
```
mukeshpanchal27, line 27: Not a blocker just for information. The `package.json` is in root folder so why we use start before it? 

```suggestion
      - 'package.json'
```

mukeshpanchal27 (approved):
Thanks @felixarntz, LGTM!

vishalkakadiya (approved):
LTGM!

.github/workflows/lint.yml
felixarntz: I took this from the Performance Lab plugin workflow which has this. Not exactly sure why, but let's keep it as is for now.

PR #47, opened 2023-03-14
Docs for xhprof

spacedmonkey:
Docs for xhprof for WordPress development environment.


xhprof/README.md
```diff
@@ -0,0 +1,17 @@
...
+### Setup with WordPress core development environment
+
+1. git clone https://github.com/WordPress/wordpress-develop.git
+2. Setup the environment using [these](https://github.com/wordPress/wordpress-develop/#development-environment-commands) instructions.
+3. Start environment, using `npm run env:start`
+4. Go to http://localhost:8889 and setup site.
+5. Stop environment, using `npm run env:stop`
+6. Copy docker-compose.override.yml into the root of wordpress-develop directory.
+7. Copy advanced-cache.php into the src/wp-content directory.
+8. Add `define( 'WP_CACHE', true );` to wp-config.php ( to enable profiler ).
```
felixarntz, line 15: Is there a way we can somehow automate this via the `docker-compose.override.yml` file? Maybe with a custom command inside it?

xhprof/docker-compose.override.yml
```diff
@@ -0,0 +1,137 @@
...
+  ##
+  # The PHP container.
+  ##
+  php:
+    image: spacedmonkey/php-xhprof:${LOCAL_PHP-latest}
+    networks:
+      - wpdevnet
+
+    environment:
+      - LOCAL_PHP_XDEBUG=${LOCAL_PHP_XDEBUG-false}
```
felixarntz, line 39: It would be great to support a environment variable like `LOCAL_PHP_XHPROF` similar to this one, so that it's easy to turn this on and off.

felixarntz (commented):
@spacedmonkey This is super neat, thanks!

Two additional suggestions for consideration.

Just for reference, here is how I previously hacked it together on my own setup: https://github.com/WordPress/wordpress-develop/commit/ed096270d817eb9850ea54e4a30662cf2d9492d8 Your solution here is _way_ better though.

xhprof/README.md
spacedmonkey: I didn't want to do this. I want developers to be able to bootstarp the profiler in how they want. I think that is a good idea for a v2 of this.

xhprof/docker-compose.override.yml
spacedmonkey: You can turn it off with WP_CACHE like this. Once this is in core, this is how I want this to work.

xhprof/README.md
felixarntz: @spacedmonkey Makes sense regarding the `WP_CACHE` constant in `wp-config.php`. But what would be the drawback of copying `advanced-cache.php` automatically for them?

xhprof/docker-compose.override.yml
felixarntz: @spacedmonkey Thanks, that makes sense to me 👍

felixarntz (commented):
@spacedmonkey Can you move all these files (i.e. the `xhprof` folder) into a new top-level directory like `tools`, so that this is in `tools/xhprof`? Otherwise we'll just get more and more folders over time.

This is the first standalone tool that is being added to this repo, and there are already PRs for other ones, so I think we should create a new top-level directory as it's clearly different from the CLI scripts and SQL queries that we have so far.

xhprof/README.md
spacedmonkey: Where would it copy it from? Have it as part of the PHP docker container? 

Maybe like this - https://github.com/WordPress/wordpress-develop/blob/6222a118c9b87ca9f5703b5055831457469627b7/docker-compose.yml#L52-L53

If we do that, maybe we should just bootstrap the file in the wp-config.php file. If that profiler is bundled with docker image, then it makes it hard to control the config.

felixarntz: @spacedmonkey Yeah, something along the lines of what you linked is what I was thinking.

I think either would work. The benefit of using `wp-config.php` would be that we don't have to "highjack" the `advanced-cache.php` file for something that is not an advanced cache. Let me know what you prefer.

tools/xhprof/README.md
```diff
@@ -0,0 +1,17 @@
...
+### Setup with WordPress core development environment
+
+1. git clone https://github.com/WordPress/wordpress-develop.git
+2. Setup the environment using [these](https://github.com/wordPress/wordpress-develop/#development-environment-commands) instructions.
+3. Start environment, using `npm run env:start`
+4. Go to http://localhost:8889 and setup site.
+5. Stop environment, using `npm run env:stop`
+6. Copy docker-compose.override.yml into the root of wordpress-develop directory.
+7. Copy advanced-cache.php into the src/wp-content directory.
+8. Add `define( 'WP_CACHE', true );` to wp-config.php ( to enable profiler ).
```
felixarntz, line 15: These two steps we could simplify. Maybe just create a completely new PHP file with the profiler code (instead of `advanced-cache.php`) and then change this documentation that they need to `require_once` the file in `wp-config.php`?

felixarntz (commented):
@spacedmonkey Apologies for the delay here.

My only outstanding feedback would be to follow up on https://github.com/GoogleChromeLabs/wpp-research/pull/47#discussion_r1137535749 and somehow make adding the profiler a less manual process. Also see related comment below.

xhprof/README.md
spacedmonkey: The profiler SDK is now part of the container. Start the profiler, is really up to you. I have provided this advanced-cache.php. Should we recommend bootstrapping in wp-config.php? 

To be clean, I would to have a profiler file accessilbe to the developer, so that can tweak it and config it.

tools/xhprof/README.md
spacedmonkey: Fixed

xhprof/README.md
spacedmonkey: I have removed the advanced-cache.php workaround and just documented use in wp-config.php include.

felixarntz (approved):
Thanks @spacedmonkey, LGTM!

felixarntz:
@joemcgill Could you please give this a review, since you have also been using XHProf quite a bit recently?

tools/xhprof/xhprof.php
```diff
@@ -0,0 +1,38 @@
...
+	$profiler = new Profiler(
+		array(
+			'save.handler.upload' => array(
+				// Use docker's internal networking to connect containers.
+				'url' => 'http://host.docker.internal:8142/run/import',
+			),
+			'profiler.flags'      => array(
+				ProfilingFlags::CPU,
+				ProfilingFlags::MEMORY,
+			),
```
joemcgill, line 28: Would be cool to add an inline doc here about adding support for not including built in PHP functions in profile runs
```suggestion
			'profiler.flags'      => array(
				ProfilingFlags::CPU,
				ProfilingFlags::MEMORY,
				// Uncomment the following line to ignore built in PHP functions.
				// ProfilingFlags::NO_BUILTINS,
			),
```

tools/xhprof/docker-compose.override.yml
```diff
@@ -0,0 +1,137 @@
...
+
+  xhgui:
+    # https://hub.docker.com/r/spacedmonkey/xhgui/tags
+    image: spacedmonkey/xhgui:latest
+    restart: always
+    environment:
+      - XHGUI_MONGO_HOSTNAME=mongo
+      - XHGUI_MONGO_DATABASE=xhprof
+    ports:
+      - "8142:80"
```
joemcgill, line 61: I think the ports you have here are fine, but it will cause conflicts if someone is trying to run both a profiler against a plugin and another one against core for some reason. To avoid conflicts we may want to consider using different ports for the Mongo DB and the XHGUI application.

joemcgill (commented):
I have a couple of comments/questions, but overall this is looking good to me.

tools/xhprof/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+[XHProf](https://www.php.net/manual/en/book.xhprof.php) is an open source PHP profile.
+
+### Setup with WordPress core development environment
+
+1. git clone https://github.com/WordPress/wordpress-develop.git
+2. Setup the environment using [these](https://github.com/wordPress/wordpress-develop/#development-environment-commands) instructions.
+3. Start environment, using `npm run env:start`
+4. Go to http://localhost:8889 and setup site.
+5. Stop environment, using `npm run env:stop`
```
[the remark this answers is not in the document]
spacedmonkey: It doesn't really. But if you haven't setup the env before, then it a good idea to get the docker images setup and the env running before the next steps.

tools/xhprof/docker-compose.override.yml
```diff
@@ -0,0 +1,137 @@
...
+    command: /bin/sh -c "envsubst '$$LOCAL_DIR' < /etc/nginx/conf.d/default.template > /etc/nginx/conf.d/default.conf && exec nginx -g 'daemon off;'"
+
+    depends_on:
+      - php
+
+  ##
+  # The PHP container.
+  ##
+  php:
+    image: spacedmonkey/php-xhprof:${LOCAL_PHP-latest}
```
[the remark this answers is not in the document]
spacedmonkey: This is not official yet and was done really quickly. If this ever get to core, then we this images will be hosted there. 

The docker images code is all here - https://github.com/spacedmonkey/wpdev-docker-images-xhprof. If anyone want to fork this and host them where ever, it is all public.

tools/xhprof/docker-compose.override.yml
spacedmonkey: This can be changed to whatever you want. This is just example. Maybe once this is in core, we can use an env variable to make this configable.

tools/xhprof/docker-compose.override.yml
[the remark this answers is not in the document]
joemcgill: Understood, and I don' think changing this needs to block an initial change here. Curious if we could host those as part of this repo or in the mono repo for WP performance? We'd still need to set up a Docker account, so I'm not sure what would be best.

joemcgill (approved):
This looks good to me. Thanks for putting it together!

PR #50, opened 2023-04-06
Add LCP minus TTFB as a metric to `benchmark-web-vitals`, using new logic for aggregate metrics

felixarntz:
This PR adds the "LCP - TTFB" metric, which is useful to assess client-side performance specifically, to the `benchmark-web-vitals` command.

Having this metric will allow us to get a more granular breakdown for client-side performance impact of a change when using this command. Note that without having this metric as part of the command, of course we could still subtract median TTFB from median LCP, but that would not necessarily give the same result, and it would most importantly be statistically incorrect. The right way is to compute this value for every single request, and then use the median of those values, which is what this PR enables.

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -206,41 +206,52 @@ async function benchmarkURL( browser, params ) {
 		}
 	} );
 
-	Object.entries( aggregateMetricsDefinition ).forEach( ( [ key, value ] ) => {
-		// Bail if any of the necessary partial metrics are not provided.
-		const partialMetrics = [ ...( value.add || [] ), ...( value.subtract || [] ) ];
-		if ( ! partialMetrics.length ) {
-			return;
-		}
-		for ( const metricKey of partialMetrics ) {
-			if ( ! metrics[ metricKey ] ) {
+	Object.entries( aggregateMetricsDefinition ).forEach(
+		( [ key, value ] ) => {
+			// Bail if any of the necessary partial metrics are not provided.
+			const partialMetrics = [
+				...( value.add || [] ),
+				...( value.subtract || [] ),
+			];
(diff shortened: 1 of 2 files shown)


felixarntz:
cc @oandregal since you asked about this a few weeks ago :)

joemcgill (approved):
Looks right to me, and works locally 👍🏻

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -131,6 +131,17 @@ async function benchmarkURL( browser, params ) {
 		},
 	};
 
+	/*
+	 * Aggregate metrics are metrics which are calculated for every request as
+	 * a combination of other metrics.
+	 */
+	const aggregateMetricsDefinition = {
+		'LCP - TTFB': {
```
westonruter, line 139: This might be more clear:

```suggestion
		'LCP w/o TTFB': {
```

Otherwise it could seem like it's reporting the TTFB part of LCP, when actually it means LCP without TTFB.

felixarntz: @westonruter Fair point. I find the "w/o" a bit odd though, not necessarily an acronym everyone is familiar with. How about "LCP-TTFB"? Maybe the lack of space makes it clear that it's actually a "minus" sign.

westonruter: Oh, I thought it would be commonly understood to mean "without". Otherwise, "LCP-TTFB" I think is a bit better.

felixarntz: Updated in https://github.com/GoogleChromeLabs/wpp-research/pull/50/commits/1723bcf799a8a67a8ab087bc8ff8a90a0355b29f.

westonruter (approved):
LGTM

PR #51, opened 2023-04-06
Add query to assess lazy-loading usage for WP sites that should have the feature

felixarntz:
This query explores how many WordPress sites >= 5.5 lazy-load any of their images. The background for that query is that we would like to get a better idea of how many WordPress sites that technically have the feature available do not use it, e.g. have it explicitly disabled.

## Query results

Row | client | urls_with_images | urls_with_loading_lazy | pct_with_loading_lazy
-- | -- | -- | -- | --
1 | desktop | 2483598 | 1546146 | 0.6225427787
2 | mobile | 3265631 | 2050648 | 0.6279484731

Based on March 2023 dataset


rviscomi (approved):

mukeshpanchal27 (approved):
Thanks @felixarntz, The PR LGTM!

PR #52, opened 2023-04-28
Prevent navigating away while benchmarking

diedexx:
When running the web vitals benchmark the click action may click a link by accident. [Puppeteer clicks the center of the element by default](https://pptr.dev/api/puppeteer.page.click).  
When this happens, Puppeteer cannot evaluate the metrics. Instead this error is thrown: `Error: Execution context was destroyed, most likely because of a navigation.` ([which is silently ignored](https://github.com/GoogleChromeLabs/wpp-research/blob/331bdc504e67111613e8ce724468187397c6a9b1/cli/commands/benchmark-web-vitals.mjs#L187)) and the results are `undefined`.
image

By giving the click action a negative offset the odds of clicking a link are greatly reduced. I suppose there could still be off-screen elements like skiplinks, but I don't expect to see anything at -500,-500.


felixarntz (approved):
Great catch, thanks @diedexx!

PR #32, opened 2023-01-10
Implement HTTP Archive query to assess usage of block themes.

felixarntz:
Fixes #9

This query identifies sites using a block theme by checking for the `<div class="wp-site-blocks">` tag which is present in all such sites, as controlled by WordPress core.

## Query results

Row | client | with_block_theme | total_wp_sites | pct_total | wp_gte_59
-- | -- | -- | -- | -- | --
1 | desktop | 7190 | 4241264 | 0.0016952493407625651 | 3393158
2 | mobile | 13074 | 5703734 | 0.0022921826298351222 | 4471610

Based on November 2022 dataset

PR #53, opened 2023-05-08
Implement query to get block theme adoption based on new HTTP Archive custom metric

felixarntz:
This query has the same purpose as the one implemented in #32, however it uses the HTTP Archive custom metric added in https://github.com/HTTPArchive/custom-metrics/pull/62, which has been available in HTTP Archive datasets starting with March 2023.

This custom metric makes the query (which was extremely slow before as it had to parse entire response bodies) notably faster (in a comparison 3:33 minutes vs 35:49 minutes).

## Query results

Row | client | total_wp_sites | with_block_theme | pct_total
-- | -- | -- | -- | --
1 | desktop | 4519655 | 31102 | 0.0068814986984626035
2 | mobile | 5829541 | 43403 | 0.00744535461711308

Based on April 2023 dataset

### Comparison with old query

For reference, to assess the new numbers and performance of the query, the query from #32 was also run again for the April 2023 dataset. Here are the results:

Row | client | with_block_theme | total_wp_sites | pct_total | wp_gte_59  
-- | -- | -- | -- | -- | -- 
1 | desktop | 12669 | 4506581 | 0.002811222077224397 | 3810151
2 | mobile | 19992 | 5825503 | 0.0034318066611587015 | 

This shows slightly lower percentages, likely because the query's approach of parsing the HTML from the response bodies is more error-prone.

### DIFF
--- a/sql/2023/05/block-theme-usage-custom-metrics.sql
+++ b/sql/2023/05/block-theme-usage-custom-metrics.sql
@@ -26,6 +26,7 @@ FROM
 WHERE
   date = "2023-04-01"
   AND technology.technology = "WordPress"
+  AND is_root_page = true
 GROUP BY
   client
 ORDER BY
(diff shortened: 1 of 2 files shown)


mukeshpanchal27 (approved):
Thanks @felixarntz, PR look good to me. I have updated table structure for old data.

Note: For old query the mobile data for `wp_gte_59` is missing.

PR #54, opened 2023-05-25
Add query to get TTFB information for localized WordPress sites

swissspidy:
This is relevant for the i18n performance efforts in WordPress core, where we want to measure TTFB improvement of localized WP sites over time.

This originally started with a breakdown by language, but for our efforts that's really not necessary, so a boolean `is_localized` flag is really enough.

-> side note: looking at passing rates per geographical region would be interesting though as another way to look at this data.

The way this flag is determined is very simple: everything not `en`/`en-us` is considered to be localized, i.e. indicating that translation loading is involved.

**Query results:**

Based on 2023-12-01 data set.

client | is_localized | ttfb_passing_rate
-- | -- | --
desktop | FALSE | 0.30744503527151634
mobile | FALSE | 0.19601053089507409  
desktop | TRUE | 0.33121590147124469  
mobile | TRUE | 0.25403273015186362

So interestingly, localized sites are already doing better today than non-localized ones.


sql/2023/06/ttfb-by-language.sql
```diff
@@ -0,0 +1,70 @@
...
+      (ttfb.start < 200, ttfb.density, 0)) / SUM(ttfb.density) AS fast,
+  SUM(
+    IF
+      (ttfb.start >= 200
+         AND ttfb.start < 1000, ttfb.density, 0)) / SUM(ttfb.density) AS avg,
+  SUM(
+    IF
+      (ttfb.start >= 1000, ttfb.density, 0)) / SUM(ttfb.density) AS slow
+FROM
+  `chrome-ux-report.all.202304`,
```
adamsilverstein, line 34: is 202304 the right table? the other tables are 05-2023.

swissspidy: There is no May table in for CrUX yet.

The HTTP Archive table is for 2023-05-01, which I assumed means it's the analyzed data from April as well.

felixarntz: @swissspidy The HTTP Archive tables always use the 1st day of the month, but that doesn't mean anything really, technically speaking those could just be called "2023-05". The pipelines are run later than that, so they contain data from _some point_ in May.

CrUX data however is only always released the month after (usually 2nd Tuesday of each month), for example May data was published on June 13 (see https://developer.chrome.com/docs/crux/release-notes/).

I'd argue for an analysis like this it doesn't matter all that much whether we look at the latest month or some time 1-2 months back, so if the latest data is not available, we can simply pick a previous month. But at this point you should be able to consistently use `2023-05-01` for both HTTP Archive and CrUX.

felixarntz, line 34: Not sure about this, do you have a reference query where you took this bit from? I'm not saying it's incorrect, but I've typically seen CWV queries use the `chrome-ux-report.materialized.device_summary` which already include `fast_ttfb`, `avg_ttfb`, and `slow_ttfb` as fields for each origin, so more efficient and less error-prone to query that rather than having logic in here to calculate it.

That approach is e.g. what the CWV Tech Report uses: https://github.com/HTTPArchive/cwv-tech-report/blob/main/sql/monthly.sql

Simple example:
```sql
CREATE TEMP FUNCTION IS_GOOD(good FLOAT64, needs_improvement FLOAT64, poor FLOAT64) RETURNS BOOL AS (
  SAFE_DIVIDE(good, good + needs_improvement + poor) >= 0.75
);

CREATE TEMP FUNCTION IS_NON_ZERO(good FLOAT64, needs_improvement FLOAT64, poor FLOAT64) RETURNS BOOL AS (
  good + needs_improvement + poor > 0
);

SELECT
  date,
  device,
  origin,
  IS_NON_ZERO(fast_ttfb, avg_ttfb, slow_ttfb) AS any_ttfb,
  IS_GOOD(fast_ttfb, avg_ttfb, slow_ttfb) AS good_ttfb
FROM
  `chrome-ux-report.materialized.device_summary`
WHERE
  date = CAST("2023-05-01" AS DATE)
```

swissspidy: Last I checked (a month ago) that table wasn‘t available yet, but I‘ll check again

sql/2023/06/ttfb-by-language.sql
```diff
@@ -0,0 +1,70 @@
...
+          AND category = 'CMS'
+          AND info != ''
+        GROUP BY
+            _TABLE_SUFFIX )
+             USING
+               (_TABLE_SUFFIX)
+    GROUP BY
+        _TABLE_SUFFIX,
+        lang,
+        url )
```
felixarntz, line 59: This is one way to do it, but since last year there's been the `httparchive.all.pages` table which makes this kind of query a lot simpler and avoids the additional `JOIN`. It's also used by CWV tech report, so maybe better to follow that example now.

Example:
```sql
SELECT
  client,
  TRIM(LOWER(JSON_VALUE(JSON_VALUE(payload, '$._almanac'), '$.html_node.lang'))) AS lang,
  page AS url
FROM
  `httparchive.all.pages`,
  UNNEST(technologies) AS technology,
  UNNEST(technology.categories) AS category,
  UNNEST(technology.info) AS version
WHERE
  date = CAST("2023-05-01" AS DATE)
  AND technology.technology = "WordPress"
  AND category = "CMS"
  AND version != ""
```

sql/2023/06/ttfb-by-language.sql
```diff
@@ -0,0 +1,70 @@
...
+         ON
+               client =
+               IF
+                 (form_factor.name = 'desktop', 'desktop', 'mobile')
+             AND CONCAT(origin, '/') = url
```
felixarntz, line 64: Just a suggestion to simplify the query and improve readability: If you adjust the field names and behavior of the subquery results to match between the two results to `JOIN`, then you can rely on the `USING` operator which is simply based on fields of the same name having the same value.

For example, you could put `CONCAT(origin, '/') AS url` as well as `IF
                 (form_factor.name = 'desktop', 'desktop', 'mobile') AS client` (or `IF
                 (device = 'desktop', 'desktop', 'mobile') AS client` if you use the alternative CrUX table I suggested above) as fields to return from the CrUX query. That way both results have `client` and `url` fields that behave the same way. Then you can use `USING (client, url)` instead of this more complicated `ON` clause.

felixarntz (commented):
@swissspidy A few suggestions for ways to simplify the query below.

sql/2023/06/ttfb-by-language.sql
swissspidy: I think I originally got it from https://github.com/HTTPArchive/almanac.httparchive.org/blob/main/sql/2022/performance/ttfb_by_cms.sql

I'm all for simplifying though 👍

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
...
+CREATE TEMP FUNCTION
+  IS_LOCALIZED(lang STRING)
+  RETURNS BOOL AS ( lang IS NOT NULL
+  AND LOWER(lang) != "en"
+  AND LOWER(lang) != "en-us" );
+WITH
+  pages AS (
+    SELECT
+      client,
+      IS_LOCALIZED(TRIM(LOWER(JSON_VALUE(JSON_VALUE(payload, '$._almanac'), '$.html_node.lang')))) AS is_localized,
```
felixarntz, line 36: Some sites use an underscore here instead of a dash. While I believe that's technically wrong, easy enough to consider, and we can assume those are still the locales those sites are actually using.
```suggestion
      IS_LOCALIZED(REPLACE(TRIM(LOWER(JSON_VALUE(JSON_VALUE(payload, '$._almanac'), '$.html_node.lang'))), '_', '-' )) AS is_localized,
```

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
...
+CREATE TEMP FUNCTION
+  IS_NON_ZERO(good FLOAT64,
+              needs_improvement FLOAT64,
+              poor FLOAT64)
+  RETURNS BOOL AS ( good + needs_improvement + poor > 0 );
+CREATE TEMP FUNCTION
+  IS_LOCALIZED(lang STRING)
+  RETURNS BOOL AS ( lang IS NOT NULL
+  AND LOWER(lang) != "en"
+  AND LOWER(lang) != "en-us" );
```
felixarntz, line 31: Nit-pick: The `LOWER` is unnecessary as `lang` is already sanitized in that way via line 36.
```suggestion
  AND lang != "en"
  AND lang != "en-us" );
```

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
...
+      `httparchive.all.pages`,
+      UNNEST(technologies) AS t
+    WHERE
+        date = '2023-10-01'
+      AND is_root_page
+      AND t.technology = 'WordPress' ),
+  devices AS (
+    SELECT
+      device,
+      origin,
```
felixarntz, line 48: Nit-pick: Those aren't used, so we can drop them to reduce data throughput for the query.

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
...
+      IS_NON_ZERO(fast_ttfb,
+                  avg_ttfb,
+                  slow_ttfb) AS any_ttfb,
+      IS_GOOD(fast_ttfb,
+              avg_ttfb,
+              slow_ttfb) AS good_ttfb
+    FROM
+      `chrome-ux-report.materialized.device_summary`
+    WHERE
+        date = CAST("2023-09-01" AS DATE)
```
felixarntz, line 61: This date should be aligned with that for the other subquery. Maybe that initially wasn't the case because CrUX data is always published a bit after HTTP Archive data, but in that case I'd still say we should always use the same month - if needed, the older one for both.

In any case, by now October data is available:
```suggestion
        date = CAST("2023-10-01" AS DATE)
```

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
...
+      AND device IN ('desktop',
+                     'tablet',
+                     'phone') )
+SELECT
+  client,
+  is_localized,
+  good_ttfb
```
felixarntz, line 68: I assume we want the TTFB passing rate here? The `good_ttfb` values isn't really that helpful on its own.
```suggestion
  SAFE_DIVIDE(COUNTIF(good_ttfb), COUNTIF(any_ttfb)) AS ttfb_passing_rate
```

sql/2023/11/ttfb-by-language.sql
```diff
@@ -0,0 +1,83 @@
+# HTTP Archive query to get TTFB information of WordPress sites by language.
```
felixarntz, line 1: Given the recent change in what this query does, let's update the description (and file name as well?). For example:
```suggestion
# HTTP Archive query to get TTFB information of WordPress sites by whether they are localized or not.
```

felixarntz (commented):
@swissspidy I left a few comments on the updated query. Let me know if you have any questions.

Once everything works, it would be great if you could update the data in the PR description too - especially as this query now does something else than it did when you originally opened the PR.

sql/2023/11/ttfb-by-language.sql
swissspidy: Just a small oversight, well spotted

swissspidy:
@felixarntz Thanks a lot for your help! This was exactly what has been missing to make this query finally correct.

sql/README.md
```diff
@@ -20,6 +20,10 @@ For additional considerations for writing BigQuery queries against HTTP Archive,
 
 ## Query index
 
+### 2023/12
```
adamsilverstein, line 23: date needs updating

adamsilverstein (approved):

westonruter (approved):

PR #55, opened 2023-06-09
XHProf: Only override modified docker compose values

joemcgill:
This edits the `docker-compose.override.yml` file so only the modified values are included. This ensures that all other values rely on the main `docker-compose.yml` values set in the core repo.

See: https://github.com/WordPress/wordpress-develop/blob/trunk/docker-compose.yml

### DIFF
--- a/tools/xhprof/docker-compose.override.yml
+++ b/tools/xhprof/docker-compose.override.yml
@@ -1,53 +1,13 @@
-version: '3.7'
-
 services:
-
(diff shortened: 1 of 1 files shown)


joemcgill:
This also removes the configuration that caused memcached to always be enabled for environments using these overrides, which I don't think was ever inteded.

spacedmonkey (approved):

PR #60, opened 2023-07-25
Pass through URL username/password to page.authenticate() for HTTP auth in benchmark-web-vitals

westonruter:
When using Local, you can expose URLs publicly via [Live Links](https://localwp.com/help-docs/local-features/live-links/). The URLs, however, are protected by HTTP Basic Auth. When using `benchmark-web-vitals` there isn't currently a way to pass the HTTP Basic Auth credentials when testing a URL, even when they are provided in the URL itself. For example:

```bash
npm run research -- benchmark-web-vitals --url http://foouser:barpass@battery-horse.localsite.io/
```

This PR ensures that username/password being provided in the URL will be passed along to the request that Puppeteer makes via the `page.authenticate()` API.

Also, when appending the `rnd` query parameter to the URL (for cache busting I presume), it instead uses the `URL` API rather than concatenating `?rnd=${ requestNum }`. This will fix testing URLs that contain query parameters.

Lastly, when there is a non-200 response when requesting a URL, the result is now logged as an error.


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -164,13 +164,28 @@ async function benchmarkURL( browser, params ) {
 			);
 
 		// Load the page.
-		const response = await page.goto(
-			`${ params.url }?rnd=${ requestNum }`,
-			{ waitUntil: 'networkidle0' }
-		);
+		const url = new URL( params.url );
+		url.searchParams.append( 'rnd', String( requestNum ) );
```
westonruter, line 168: Question: Rather than using `requestNum` here, shouldn't an actual random number be used or a UUID? Otherwise, when repeated calls are made there will be duplicate requests to `/?rnd=0` which could be cached and would produce misleading results.

felixarntz: Good point, I think that makes sense.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -164,13 +164,28 @@ async function benchmarkURL( browser, params ) {
...
+			waitUntil: 'networkidle0',
+		} );
 		await page.addScriptTag( { content: scriptTag, type: 'module' } );
 
 		if ( response.status() !== 200 ) {
+			log(
+				formats.error(
+					`Request to ${ url } returned status ${ response.status() }.`
+				)
+			);
```
felixarntz, line 188: Not sure we want the logging here, at least not unconditionally. There's always a chance that a few requests fail, which is why at the end the completion rate is shown, and logging every failed request would clutter the CLI output. If we want to add logging for this, I think it should only be in some `--verbose` mode or something like that.

felixarntz (commented):
Thanks @westonruter, this mostly looks good. Left a few comments.

cli/commands/benchmark-web-vitals.mjs
westonruter: Yeah, it's just that if there is a typo in the URL or something and the completion rate is 0% then there's no indication of what went wrong at present. So then you have to re-request the URL, e.g. via `curl`, to debug further.

westonruter, line 188: I'll just remove for now.

```suggestion
```

felixarntz: Maybe we store that error in a variable and only log it if at the end there's a 0% success rate, indicating that something in the actual command usage was wrong?

westonruter: Yeah, that'd work. We can explore that in another PR.

cli/commands/benchmark-web-vitals.mjs
westonruter: How about just this:

```suggestion
		url.searchParams.append( 'rnd', String( Math.random() ) );
```

felixarntz (approved):

PR #61, opened 2023-07-26
Update dependencies & Node version

swissspidy:
Node 14 is EOL.

Node 18 is LTS.

Node has had built-in fetch for a while now.

Technically files don’t need to have an `.mjs` file extension, but it doesn’t really matter.


westonruter (approved):
LGTM

westonruter:
When running `benchmark-web-vitals` I get this warning:

```
  Puppeteer old Headless deprecation warning:
    In the near feature `headless: true` will default to the new Headless mode
    for Chrome instead of the old Headless implementation. For more
    information, please see https://developer.chrome.com/articles/new-headless/.
    Consider opting in early by passing `headless: "new"` to `puppeteer.launch()`
    If you encounter any bugs, please report them to https://github.com/puppeteer/puppeteer/issues/new/choose.
```

I think this PR should also add `{headless: "new"}` to `puppeteer.launch()` as advised.

swissspidy:
Sure, why not, feel free to amend the PR.
I don‘t expect it to break anything. Otherwise we can migrate later, no problem. It‘s not urgent.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -78,7 +78,7 @@ export async function handler( opt ) {
 	const { number: amount } = opt;
 	const results = [];
 
-	const browser = await puppeteer.launch();
+	const browser = await puppeteer.launch({ headless: "new" });
```
westonruter, line 81:
```suggestion
	const browser = await puppeteer.launch({ headless: 'new' });
```

joemcgill (approved):
👍🏻 Thanks for handling this

PR #63, opened 2023-08-01
Add queries to determine frequency of themes/plugins adding blocking/async/defer scripts in head/footer

westonruter:
# WordPress theme/plugin script placements (whether blocking/async/defer in head/footer)

Query: `theme-plugin-script-placements.sql`

Results for July 2023:

placement	| num_scripts
--|--
blocking_in_footer	| 87,096,709
blocking_in_head	| 16,193,819
defer_in_footer	| 5,999,869
async_in_footer	| 3,291,749
defer_in_head	| 1,068,317
async_in_head	| 941,814

Duration: 4 min 26 sec

# Blocking in head scripts from themes and plugins

Query: `blocking-in-head-scripts-from-themes-and-plugins.sql`

Note the rows returned by this query include not only the counts of the specific scripts but also the counts for all scripts from a given theme/plugin (e.g. both `plugin:revslider:public/assets/js/rs6.min.js` and `plugin:revslider`).

Top 25 results from July 2023:

source | source_count
-- | --
plugin:revslider | 909,948
plugin:woocommerce | 559,495
plugin:elementor | 452,980
plugin:elementor:assets/lib/font-awesome/js/v4-shims.min.js | 442,141
plugin:LayerSlider | 410,705
plugin:pixelyoursite | 370,128
plugin:photo-gallery | 367,291
plugin:cookie-law-info | 361,501
plugin:google-analytics-for-wordpress | 360,082
plugin:google-analytics-for-wordpress:assets/js/frontend-gtag.min.js | 330,632
plugin:Ultimate_VC_Addons | 268,471
plugin:revslider:public/assets/js/rs6.min.js | 265,917
plugin:woocommerce:assets/js/jquery-blockui/jquery.blockUI.min.js | 245,665
plugin:woocommerce:assets/js/frontend/add-to-cart.min.js | 237,492
plugin:js_composer | 230,243
plugin:js_composer:assets/js/vendors/woocommerce-add-to-cart.js | 227,278
plugin:cookie-law-info:legacy/public/js/cookie-law-info-public.js | 226,390
plugin:revslider:public/assets/js/rbtools.min.js | 212,473
plugin:sitepress-multilingual-cms | 192,425
plugin:revslider:public/assets/js/jquery.themepunch.revolution.min.js | 179,701
plugin:revslider:public/assets/js/jquery.themepunch.tools.min.js | 170,372
plugin:responsive-lightbox | 166,425
plugin:download-manager | 136,044
plugin:wp-user-avatar | 134,159
plugin:pixelyoursite:dist/scripts/jquery.bind-first-0.2.3.min.js | 128,739

Duration: 3 min 20 sec

### DIFF
--- a/sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
+++ b/sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
@@ -1,0 +1,77 @@
+CREATE TEMP FUNCTION GET_BLOCKING_HEAD_SOURCES (custom_metrics STRING) RETURNS ARRAY<STRING> LANGUAGE js AS '''
+
+const sourceRegExp = new RegExp( '/wp-content/(?<type>plugin|theme)s/(?<slug>[^/]+)/(?<path>[^\?]+)' );
+
+/**
+ * Get slug of extension prefixed by theme/plugin from URL.
+ *
+ * @param {string} src Script URL.
+ * @return {?{type: "plugin"|"theme", slug: string, path: string}} Source info if matched.
+ */
+function getSource(src) {
+  const matches = src.match( sourceRegExp );
+  if (matches) {
+    return matches.groups;
+  }
+  return null;
+}
+
+/**
+ * Get script sources for scripts in the head which are blocking (not async nor defer).
+ *
+ * @param {object} data
+ * @param {object} data.cms
+ * @param {object} data.cms.wordpress
+ * @param {Array<{src: string, intended_strategy: string, async: boolean, defer: boolean, in_footer: boolean}>} data.cms.wordpress.scripts
+ * @return {Array} Sources.
+ */
+function getBlockingHeadScriptSources(data) {
+  const sources = [];
+  for ( const script of data.cms.wordpress.scripts ) {
+
+    const source = getSource(script.src);
+    if (!source) {
+      continue;
+    }
+
+    // Blocking script in head only.
+    if (!script.in_footer && !script.async && !script.defer) {
+      sources.push( [ source.type, source.slug ].join( ':' ) );
+      sources.push( [ source.type, source.slug, source.path ].join( ':' ) );
+    }
+  }
+  return sources;
+}
+
+const sources = [];
+try {
+  const data = JSON.parse(custom_metrics);
+  sources.push(...getBlockingHeadScriptSources(data));
+} catch (e) {}
+return sources;
+''';
+
+WITH all_sources AS (
+  SELECT
+    GET_BLOCKING_HEAD_SOURCES(custom_metrics) AS sources,
+  FROM
+    `httparchive.all.pages` /*TABLESAMPLE SYSTEM (1 PERCENT)*/,
+    UNNEST(technologies) AS technology
+  WHERE
+  date = CAST("2023-07-01" AS DATE)
+  AND technology.technology = "WordPress"
+  AND is_root_page = TRUE
+)
+
+SELECT
+  source,
+  COUNT(source) AS source_count
+FROM
+  all_sources,
+  UNNEST(sources) AS source
+GROUP BY
+  source
+HAVING
+  source_count >= 500
+ORDER BY
+  source_count DESC
--- a/sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
+++ b/sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
@@ -1,3 +1,20 @@
+# HTTP Archive query to get counts of theme/plugin scripts blocking in head.
+#
+# WPP Research, Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/63
 CREATE TEMP FUNCTION GET_BLOCKING_HEAD_SOURCES (custom_metrics STRING) RETURNS ARRAY<STRING> LANGUAGE js AS '''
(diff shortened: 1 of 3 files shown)


sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
```diff
@@ -0,0 +1,77 @@
...
+SELECT
+  source,
+  COUNT(source) AS source_count
+FROM
+  all_sources,
+  UNNEST(sources) AS source
+GROUP BY
+  source
+HAVING
+  source_count >= 500
```
felixarntz, line 75: Why is this limitation here? How many results make it above this threshold? If it's too many, we may even want to set it higher?

sql/2023/08/theme-plugin-script-placements.sql
```diff
@@ -0,0 +1,63 @@
...
+  SELECT
+    GET_SCRIPT_PLACEMENTS(custom_metrics) AS placements,
+  FROM
+    `httparchive.all.pages` /*TABLESAMPLE SYSTEM (1 PERCENT)*/,
+  UNNEST(technologies) AS technology
+WHERE
+  date = CAST("2023-07-01" AS DATE)
+  AND technology.technology = "WordPress"
+  AND is_root_page = TRUE
+  )
```
felixarntz, line 52: Indentation is off here :)
```suggestion
    UNNEST(technologies) AS technology
  WHERE
    date = CAST("2023-07-01" AS DATE)
    AND technology.technology = "WordPress"
    AND is_root_page = TRUE
)
```

sql/2023/08/theme-plugin-script-placements.sql
```diff
@@ -0,0 +1,63 @@
...
+  placements.push(...getScriptPlacements(data));
+} catch (e) {}
+return placements;
+''';
+
+WITH all_placements AS (
+  SELECT
+    GET_SCRIPT_PLACEMENTS(custom_metrics) AS placements,
+  FROM
+    `httparchive.all.pages` /*TABLESAMPLE SYSTEM (1 PERCENT)*/,
```
felixarntz, line 46: Probably best to remove the comment here for merging the query.
```suggestion
    `httparchive.all.pages`,
```

sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
```diff
@@ -0,0 +1,77 @@
...
+  sources.push(...getBlockingHeadScriptSources(data));
+} catch (e) {}
+return sources;
+''';
+
+WITH all_sources AS (
+  SELECT
+    GET_BLOCKING_HEAD_SOURCES(custom_metrics) AS sources,
+  FROM
+    `httparchive.all.pages` /*TABLESAMPLE SYSTEM (1 PERCENT)*/,
```
felixarntz, line 58: See my other comment:
```suggestion
    `httparchive.all.pages`,
```

felixarntz (commented):
Thanks @westonruter, this looks great.

For the general script placements query, I couldn't tell from the code review whether the approach of gathering values into a placements array and then counting would work correctly, so I implemented a simple query to validate the result is the same. This query basically computes only the value for `async_in_footer`, by counting directly in JS and then using a `SUM`:
```sql
CREATE TEMP FUNCTION GET_ASYNC_IN_FOOTER_COUNT (custom_metrics STRING) RETURNS INT64 LANGUAGE js AS '''

const sourceRegExp = new RegExp('/wp-content/(plugin|theme)s/([^/]+)/');

function getAsyncInFooterCount(data) {
  let counter = 0;
  for (const script of data.cms.wordpress.scripts) {
    if (!sourceRegExp.test(script.src)) {
      continue;
    }

    if (script.async && script.in_footer) {
      counter++;
    }
  }
  return counter;
}

try {
  const data = JSON.parse(custom_metrics);
  return getAsyncInFooterCount(data);
} catch (e) {}
  return 0;
''';

SELECT
  SUM(GET_ASYNC_IN_FOOTER_COUNT(custom_metrics)) AS async_in_footer
FROM
  `httparchive.all.pages` /*TABLESAMPLE SYSTEM (1 PERCENT)*/,
UNNEST(technologies) AS technology
WHERE
  date = CAST("2023-07-01" AS DATE)
  AND technology.technology = "WordPress"
  AND is_root_page = TRUE
```

The result is the same number you have posted on the PR description, so that gives confidence. I think that technically the query would be easier to understand if we were doing all the counts in JS and return e.g. an object/struct like `{ placement: "async_in_footer", number: 3291749 }` since then we would simply need to use `SUM` in the aggregation and we would be able to avoid working with arrays in BigQuery, which can sometimes be complicated and lead to odd side effects when unnesting: For example, unnesting can lead to additional rows appearing, which can skew counts; not the case here because the results are being grouped, but I've ran into the problem before.

Not a necessity, but IMO it would be a nice enhancement to simplify the query.

Other than that, only minor nitpicks below.

sql/2023/08/blocking-in-head-scripts-from-themes-and-plugins.sql
westonruter: The tail is verrrrrrrrrry long at 267K rows. This brings the row count down to ~5K which is much more reasonable to work with in a Sheet, but true it could be higher. If we set it to:

```suggestion
  source_count >= 10000
```

Then the row count would shrink down to 332.

sql/2023/08/theme-plugin-script-placements.sql
westonruter: I found it useful to keep in order to quickly remove to test. But I don't feel strongly.

adamsilverstein (commented):
Nice!

felixarntz (changes requested):
@westonruter Code-wise LGTM now, other than the threshold conversation, but can you please add the license header found in the other SQL queries, including a link to this PR for the results?

felixarntz (approved):

adamsilverstein (approved):
Fantastic!

westonruter:
> plugin:woocommerce:assets/js/jquery-blockui/jquery.blockUI.min.js	245,665
> plugin:woocommerce:assets/js/frontend/add-to-cart.min.js	237,492

Humm. This doesn't seem right. I just checked and WooCommerce scripts are enqueued in the footer. I don't see this as being a recent change either:

https://github.com/woocommerce/woocommerce/blob/ff51b07da317cd535958e60e7f8be9ca2dfb1c23/plugins/woocommerce/includes/class-wc-frontend-scripts.php#L191-L320

PR #64, opened 2023-08-04
Add query to gather counts for types of inline scripts

westonruter:
This allows us to determine the relative frequency of the types of inline scripts in `WP_Scripts`, namely before, after, extra (`wp_localize_script()`), and translations.

Output for July 2023:

Type | Count | Percentage
-- | --: | --:
extra | 25,082,363,691 | 65.26%
before | 8,184,494,232 | 21.29%
after | 3,782,332,149 | 9.84%
translations | 1,384,873,280 | 3.60%
**Total:** | 38,434,063,352 |


sql/2023/08/inline-script-types.sql
```diff
@@ -0,0 +1,70 @@
...
+ * @param {Array<{after_script_size: number, before_script_size: number, extra_script_size: number, translations_script_size: number}>} data.cms.wordpress.scripts
+ * @return {Array} Script types.
+ */
+function getScriptTypes(data) {
+  const scriptTypes = [];
+  for (const script of data.cms.wordpress.scripts) {
+    scriptTypes.push( (new Array( script.after_script_size )).fill( 'after' ) );
+    scriptTypes.push( (new Array( script.before_script_size )).fill( 'before' ) );
+    scriptTypes.push( (new Array( script.extra_script_size )).fill( 'extra' ) );
+    scriptTypes.push( (new Array( script.translations_script_size )).fill( 'translations' ) );
```
westonruter, line 35: Since these metrics are already numbers, it does seem like this should rather be implemented to return a `STRUCT` and utilize `SUM()` rather than `COUNT()` in the query.

PR #68, opened 2023-08-10
Add WP version comparison script

swissspidy:
Ports over the GitHub action and the manual Bash script from https://github.com/swissspidy/compare-wp-performance

The individual `package.json` files are unfortunately necessary right now because of https://github.com/WordPress/gutenberg/issues/51419


tools/compare-wp-performance/run.sh
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env bash
```
felixarntz, line 1: Nit-pick, but it would be great if you could make this script work regardless from where you run it. Right now it only works if you first `cd` into the `tools/compare-wp-performance` directory. Running `./tools/compare-wp-performance/run.sh` from the root directory gives errors due to the locations.

tools/compare-wp-performance/new/package.json
```diff
@@ -0,0 +1,24 @@
...
+	},
+	"author": "swissspidy",
+	"license": "Apache-2.0",
+	"bugs": {
+		"url": "https://github.com/swissspidy/compare-wp-performance/issues"
+	},
+	"homepage": "https://github.com/swissspidy/compare-wp-performance#readme",
+	"type": "module",
+	"devDependencies": {
+		"@wordpress/env": "^8.4.0",
```
felixarntz, line 17: Couldn't we alternatively use an older version that doesn't have the bug and thus doesn't require the patch and extra `package.json` workaround?

tools/compare-wp-performance/old/package.json
```diff
@@ -0,0 +1,24 @@
...
+	"type": "module",
+	"devDependencies": {
+		"@wordpress/env": "^8.4.0",
+		"patch-package": "^7.0.0"
+	},
+	"scripts": {
+		"postinstall": "patch-package",
+		"wp-env": "wp-env"
+	}
+}
```
felixarntz, line 24: Can you clarify why we need the `package.json` twice, one for the `new` and one for the `old` folder? Couldn't those share the dependencies between a single `package.json`?

tools/compare-wp-performance/scripts/results.js
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env node
+
+import { existsSync } from 'node:fs';
```
felixarntz, line 3: I'm seeing errors when running the process:
```
/Users/felixarntz/Development/wpp-research/tools/compare-wp-performance/scripts/results.js:3
import { existsSync } from 'node:fs';
^^^^^^

SyntaxError: Cannot use import statement outside a module
```

felixarntz (commented):
Thanks @swissspidy, this looks great overall! A few minor comments. I wasn't able to complete the process locally due to an error.

tools/compare-wp-performance/old/package.json
swissspidy: A locally installed wp-env only works in the current folder as that‘s where it‘s looking for the config file

tools/compare-wp-performance/new/package.json
swissspidy: I will need to check what‘s possible / where the bug was introduced and if there are features we can live wirhout or not.

swissspidy: The bug has since been resolved, so we can use a newer version without a patch, but it still requires 2 package.json files. Or we leverage a global install.

swissspidy:
Yeah I'll need to revisit this and merge in the latest changes. But holding off a bit now because it's not urgent (for 6.4 we can still use the existing repo) and there might be significant changes needed when tackling all the variance issues.

PR #72, opened 2023-09-19
Add command to analyze a URL for loading optimization issues

westonruter:
In analyzing URLs from the [WordPress 6.3 lazy-loading and fetchpriority usage analysis in the field](https://colab.sandbox.google.com/drive/1lbTFTy1IoifpqkynKuualVjrLVSj1i0r#scrollTo=BkDefWwDrQHD) Colab which have `fetchpriority` and lazy-loading issues, I found I was manually encountering the same issues over and over again. To avoid having to manually perform these checks and to automate that we can fix the issues and add regression testing in the future, I'm adding a command that summarizes what the issues are (if any). This will be particularly useful in the development of https://github.com/WordPress/performance/issues/805.

For example:

```
$ npm run --silent research analyze-loading-optimization -- -u https://wordpress.org/ -o table
╔══════════════════════════════╤════════╤═════════╗
║ field                        │ mobile │ desktop ║
╟──────────────────────────────┼────────┼─────────╢
║ lcpMetric                    │ 991.5  │ 620.5   ║
╟──────────────────────────────┼────────┼─────────╢
║ lcpElement                   │ H1     │ IMG     ║
╟──────────────────────────────┼────────┼─────────╢
║ lcpElementIsLazyLoaded       │ false  │ false   ║
╟──────────────────────────────┼────────┼─────────╢
║ lcpImageMissingFetchPriority │ false  │ true    ║
╟──────────────────────────────┼────────┼─────────╢
║ fetchPriorityCount           │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ fetchPriorityInsideViewport  │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ fetchPriorityOutsideViewport │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ lazyLoadableCount            │ 14     │ 14      ║
╟──────────────────────────────┼────────┼─────────╢
║ lazyLoadedInsideViewport     │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ lazyLoadedOutsideViewport    │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ eagerLoadedInsideViewport    │ 0      │ 0       ║
╟──────────────────────────────┼────────┼─────────╢
║ eagerLoadedOutsideViewport   │ 14     │ 14      ║
╟──────────────────────────────┼────────┼─────────╢
║ errors                       │ 14     │ 15      ║
╚══════════════════════════════╧════════╧═════════╝
```

An analysis is done for the URL emulating both mobile and desktop in viewports and the user agent. The results are presented side-by-side for comparison. The last row is a count of the number of optimization encountered when analyzing the site. In the example above for wordpress.org, the error count for mobile is one greater than that for desktop because the LCP element is an image and it doesn't have `fetchpriority=high`. (Note to self: Even when a `h1` is the LCP element, it still probably makes sense to add `fetchpriority=high` to the largest image in the viewport.)

The error conditions are as follows:

Code | Description
-----|------------
`LCP_IMAGE_MISSING_FETCHPRIORITY` | The LCP image is missing `fetchpriority=high` (and the image doesn't have a `data:` URL).
`LCP_ELEMENT_IS_LAZY_LOADED` | The LCP element (`img` or `iframe`) has `loading=lazy`.
`LAZY_LOADED_ELEMENT_IN_INITIAL_VIEWPORT` | A lazy-loaded element (`img` or `iframe`) is in the initial viewport. (May occur many times.)
`FETCHPRIORITY_OUTSIDE_VIEWPORT` | An image with `fetchpriority=high` is outside the initial viewport. (May occur multiple times.)
`EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT` |  A eager-loaded element (`img` or `iframe`) is outside the initial viewport.

In addition to the `table` output format, the same data can be presented in CSV format by providing the `csv` value (same as for other commands).

Two additional formats are supported. The first is `json` which dumps the raw data, including the underlying error codes instead of just an error count:

```json
{
    "url": "https://wordpress.org/",
    "deviceAnalyses": {
        "mobile": {
            "lcpMetric": 1021.2000000476837,
            "lcpElement": "H1",
            "lcpElementIsLazyLoaded": false,
            "lcpImageMissingFetchPriority": false,
            "fetchPriorityCount": 0,
            "fetchPriorityInsideViewport": 0,
            "fetchPriorityOutsideViewport": 0,
            "lazyLoadableCount": 14,
            "lazyLoadedInsideViewport": 0,
            "lazyLoadedOutsideViewport": 0,
            "eagerLoadedInsideViewport": 0,
            "eagerLoadedOutsideViewport": 14,
            "errors": [
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT"
            ]
        },
        "desktop": {
            "lcpMetric": 523.6999999880791,
            "lcpElement": "IMG",
            "lcpElementIsLazyLoaded": false,
            "lcpImageMissingFetchPriority": true,
            "fetchPriorityCount": 0,
            "fetchPriorityInsideViewport": 0,
            "fetchPriorityOutsideViewport": 0,
            "lazyLoadableCount": 14,
            "lazyLoadedInsideViewport": 0,
            "lazyLoadedOutsideViewport": 0,
            "eagerLoadedInsideViewport": 0,
            "eagerLoadedOutsideViewport": 14,
            "errors": [
                "LCP_IMAGE_MISSING_FETCHPRIORITY",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT",
                "EAGER_LOADED_ELEMENT_OUTSIDE_INITIAL_VIEWPORT"
            ]
        }
    }
}
```

Lastly, there is a `csv-oneline` format which is useful for collating results for a set of URLs. For example, given a file `urls.txt`:

```
https://wordpress.org/
https://make.wordpress.org/
https://wptavern.com/
```

And given a Bash script `batch-analyze-loading-optimization.sh`:

```bash
#!/bin/bash

url_count=0
while read url; do
	echo "$url_count. $url" > /dev/stderr
	if npm run --silent research analyze-loading-optimization -- -u "$url" -o csv-oneline | tail -n $( [[ $url_count = 0 ]] && echo 2 || echo 1 ); then
		url_count=$(($url_count + 1 ))
	fi
done
```

Running this command:

```bash
cat urls.txt | ./batch-analyze-loading-optimization.sh > analyses.csv
```

Results in an `analyses.csv` file that contains:

```
url,mobile:lcpMetric,mobile:lcpElement,mobile:lcpElementIsLazyLoaded,mobile:lcpImageMissingFetchPriority,mobile:fetchPriorityCount,mobile:fetchPriorityInsideViewport,mobile:fetchPriorityOutsideViewport,mobile:lazyLoadableCount,mobile:lazyLoadedInsideViewport,mobile:lazyLoadedOutsideViewport,mobile:eagerLoadedInsideViewport,mobile:eagerLoadedOutsideViewport,mobile:errors,desktop:lcpMetric,desktop:lcpElement,desktop:lcpElementIsLazyLoaded,desktop:lcpImageMissingFetchPriority,desktop:fetchPriorityCount,desktop:fetchPriorityInsideViewport,desktop:fetchPriorityOutsideViewport,desktop:lazyLoadableCount,desktop:lazyLoadedInsideViewport,desktop:lazyLoadedOutsideViewport,desktop:eagerLoadedInsideViewport,desktop:eagerLoadedOutsideViewport,desktop:errors
https://wordpress.org/,917.4,IMG,false,true,0,0,0,14,0,0,0,14,15,659.8,IMG,false,true,0,0,0,14,0,0,0,14,15
https://make.wordpress.org/,926.6,IMG,false,false,1,1,0,3,0,0,1,2,2,650.1,IMG,false,false,1,1,0,3,0,0,1,2,2
https://wptavern.com/,996.5,IMG,true,true,0,0,0,12,2,8,0,2,6,722.5,IMG,true,true,0,0,0,12,2,8,0,2,6
```

This can then be, for example, pasted into a Google Sheet for further analysis.


cli/commands/analyze-loading-optimization.mjs
```diff
@@ -0,0 +1,527 @@
...
+
+	/** @type {DeviceAnalysis} */
+	const finalAnalysis = await page.evaluate( ( webVitalsLcpGlobal ) => {
+		/* eslint-env browser */
+
+		/**
+		 * Checks whether an element is in the viewport.
+		 *
+		 * @todo This should return a percentage of how much the element is in the viewport.
+		 * @todo We should also factor in how much an element is outside the viewport. If there is an eager loaded image that is just outside the viewport, this should be OK.
```
westonruter, line 329: See also how HTTP Archive's `lcp_elem_stats` includes a `cover90viewport` boolean.

cli/commands/analyze-loading-optimization.mjs
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * @param {Object} opt
+ */
+export async function handler( opt ) {
+	const params = getParamsFromOptions( opt );
+	const browser = await puppeteer.launch( {
+		headless: 'new',
+		// TODO: Command is not working when opening in non-headless mode. The LCP never fires.
+		// headless: false, devtools: true
```
felixarntz, line 128: Why does that work when using `benchmark-web-vitals` though? I think it may be good to figure that out so that we can use headless.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -164,7 +164,7 @@ export async function handler( opt ) {
 	for await ( const url of getURLs( opt ) ) {
 		const { completeRequests, metrics } = await benchmarkURL(
 			browser,
-			params
+			params // TODO: This should be passing the url and not params.url.
```
felixarntz, line 167: Can you clarify why? Further down in `benchmarkURL()` you added a comment that the viewport should be configurable, so I'd argue that alone would be a good reason to keep using `params`.

cli/commands/analyze-loading-optimization.mjs
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * External dependencies
+ */
+import puppeteer, { Browser } from 'puppeteer';
+import round from 'lodash-es/round.js';
+
+/* eslint-disable jsdoc/valid-types */
+/** @typedef {import("web-vitals").LCPMetricWithAttribution} LCPMetricWithAttribution */
+/* eslint-enable jsdoc/valid-types */
```
felixarntz, line 27: Do we really need this? As far as I can tell this is the only reason that you added the `web-vitals` dependency to `package.json`, which to me is not useful enough. Especially since then the `jsdoc/valid-types` rule gets disabled for the line anyway.

Can't we work without the concrete `@typedef` here, given we don't actually need the `web-vitals` dependency (since it's only injected into the browser from CDN anyway)?

package.json
```diff
@@ -14,7 +14,8 @@
     "csv-stringify": "^6.4.0",
     "lodash-es": "4.17.21",
     "puppeteer": "^20",
-    "table": "^6.8.0"
+    "table": "^6.8.0",
+    "web-vitals": "3.5.0"
```
felixarntz, line 18: See above, I'm not convinced about this change.

felixarntz (changes requested):
@westonruter I left a few points of feedback below.

Additionally, could you please add some documentation with basic examples for the command to `cli/README.md`?

cli/commands/analyze-loading-optimization.mjs
westonruter: Actually, the situation is the reverse. It _does_ work with `headless`. It just doesn't work when headful. So it works the same as with `benchmark-web-vitals`.

cli/commands/benchmark-web-vitals.mjs
westonruter: I added this because `params` contains either `url` or `file`. The list of URLs is obtained from `getURLs()`. It doesn't seem to be accounting for the case where the URLs are supplied via `file`. It is assuming that `getURLs( opt )[0]` is always the same as `params.url`, as I understand.

cli/commands/analyze-loading-optimization.mjs
westonruter: Yes, the benefit of the `web-vitals` dependency is for the typing.

 For example, when we are working with `webVitalsLCP`:

image

Then the editor can know that `webVitalsLCP.attribution` is is an `LCPAttribution` object:

image

That its `lcpEntry` property is a `LargestContentfulPaint` object:

image

And that it has an `element` property which is an `Element` object:

image

We get autocompletion as well:

image

Without the `web-vitals` types, then the editor doesn't know what properties the object has:

image

cli/commands/benchmark-web-vitals.mjs
westonruter: For example, if I do this:

```
npm run research benchmark-web-vitals -- --file urls.txt 
```

Where `urls.txt` contains:

```
https://wordpress.org/
https://make.wordpress.org/
https://wptavern.com/
```

I get an error:

```
> research
> ./cli/run.mjs benchmark-web-vitals --file urls.txt

TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:405:5)
    at new URL (node:internal/url:637:13)
    at benchmarkURL (file:///home/westonruter/repos/wpp-research/cli/commands/benchmark-web-vitals.mjs:253:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async handler (file:///home/westonruter/repos/wpp-research/cli/commands/benchmark-web-vitals.mjs:165:41)
    at async Command.<anonymous> (file:///home/westonruter/repos/wpp-research/cli/run.mjs:65:4) {
  input: 'undefined',
  code: 'ERR_INVALID_URL'
}
```

This is is erroring here:

```js
const url = new URL( params.url );
```

Because the URLs aren't being provided via `--url` but rather via `--file`.

westonruter: Fixed in 075bb5f:

```
$ npm run --silent research benchmark-web-vitals -- --file urls.txt 
╔═══════════════════╤════════════════════════╤═════════════════════════════╤═══════════════════════╗
║ URL               │ https://wordpress.org/ │ https://make.wordpress.org/ │ https://wptavern.com/ ║
╟───────────────────┼────────────────────────┼─────────────────────────────┼───────────────────────╢
║ Success Rate      │ 100%                   │ 100%                        │ 100%                  ║
╟───────────────────┼────────────────────────┼─────────────────────────────┼───────────────────────╢
║ FCP (median)      │ 1081.7                 │ 806.1                       │ 1515.1                ║
╟───────────────────┼────────────────────────┼─────────────────────────────┼───────────────────────╢
║ LCP (median)      │ 1081.7                 │ 806.1                       │ 1523.4                ║
╟───────────────────┼────────────────────────┼─────────────────────────────┼───────────────────────╢
║ TTFB (median)     │ 626.7                  │ 470.9                       │ 997.3                 ║
╟───────────────────┼────────────────────────┼─────────────────────────────┼───────────────────────╢
║ LCP-TTFB (median) │ 455                    │ 335.2                       │ 526.1                 ║
╚═══════════════════╧════════════════════════╧═════════════════════════════╧═══════════════════════╝
```

felixarntz: Thanks, now I get it. Great catch!

felixarntz (approved):
Thanks @westonruter, LGTM!

swissspidy (approved):

PR #78, opened 2023-10-26
Enhance `benchmark-web-vitals` with specific metric filtering and Server-Timing support

felixarntz:
This PR adds two features to the `benchmark-web-vitals` command:
1. A `--metrics` (`-m`) parameter that allows including specific metrics. You could use it to e.g. only get "TTFB" returned. The default metrics if the parameter is not provided remain the same ones as they have been so far ("FCP", "LCP", "TTFB", "LCP-TTFB").
2. Support for including specific Server-Timing metrics in the results, via the new parameter. You can pass any metric name prefixed with "ST:" (e.g. "ST:wp-total"), and it will be included in the data. Obviously it'll only have values if the server response includes the relevant header.

While 1. is certainly a nice little convenience enhancement, 2. may seem a bit questionable at first glance since we already have the `benchmark-server-timing` command, which is much more efficient for that. And for use-cases where only Server-Timing values are needed, or where the reduced variance of using `curl` for Server-Timing metrics is of particular importance, `benchmark-server-timing` continues to be the better choice. However, optionally supporting Server-Timing metrics to be included in `benchmark-web-vitals` allows to get these metrics from a browser request together with e.g. certain Web Vitals. An example use-case (and the reason I worked on this in the first place) is to compare `wp-total` as returned by the server with actual TTFB.

For examples using these two new features, see the updated docs in the readme file.

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -97,6 +97,16 @@ export const options = [
  */
 
 /**
+ * @typedef {Object} MetricsDefinitionEntry
+ * @property {string}    type     Either 'webVitals', 'serverTiming', or 'aggregate'.
+ * @property {?string}   listen   Which event to listen to (only relevant for type 'webVitals').
+ * @property {?string}   global   Which JS global to find the metric in (only relevant for type 'webVitals').
+ * @property {?string[]} add      Which other metrics to add (only relevant for type 'aggregate').
+ * @property {?string[]} subtract Which other metrics to subtract (only relevant for type 'aggregate').
+ * @property {?string}   name     Name of the Server-Timing metric (only relevant for type 'serverTiming').
+ */
+
+/**
  * @param {Object}                opt
  * @param {?string}               opt.url
  * @param {string|number}         opt.number
@@ -168,7 +178,7 @@ function getParamsFromOptions( opt ) {
 
 /**
  * @param {string[]} metrics
- * @return {Object} Metrics definition, keyed my metric identifier.
+ * @return {Object<string, MetricsDefinitionEntry>} Metrics definition, keyed by metric identifier.
  */
 function getMetricsDefinition( metrics ) {
 	/*
@@ -201,9 +211,12 @@ function getMetricsDefinition( metrics ) {
 		},
 	};
 
-	/*
+	/**
 	 * Server-Timing metrics can have any name, so for those a generic definition creator is used.
 	 * These metrics must be prefixed with "ST:", see below.
+	 *
+	 * @param {string} metric
+	 * @return {MetricsDefinitionEntry} Server-Timing metrics definition entry.
 	 */
 	const getServerTimingDefinition = ( metric ) => {
 		return {
@@ -290,10 +303,10 @@ export async function handler( opt ) {
 }
 
 /**
- * @param {string}  url
- * @param {Browser} browser
- * @param {Object}  metricsDefinition
- * @param {Params}  params
+ * @param {string}                 url
+ * @param {Browser}                browser
+ * @param {MetricsDefinitionEntry} metricsDefinition
+ * @param {Params}                 params
  * @return {Promise<{completeRequests: number, metrics: {}}>} Results
  */
 async function benchmarkURL( url, browser, metricsDefinition, params ) {
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -222,10 +222,15 @@ function getMetricsDefinition( metrics ) {
 		return {
 			type: 'serverTiming',
 			name: metric,
+			listen: null,
+			global: null,
+			add: null,
+			subtract: null,
 		};
 	};
 
 	// Set up object with the requested metrics, and store aggregate metrics in a list.
+	/** @type {Object<string, MetricsDefinitionEntry>} */
 	const metricsDefinition = {};
 	const aggregates = [];
 	for ( const metric of metrics ) {
@@ -400,10 +405,8 @@ async function benchmarkURL( url, browser, metricsDefinition, params ) {
 							offset: { x: -500, y: -500 },
 						} );
 						// Get the metric value from the global.
-						/** @type {number} */
-						const metric = await page.evaluate(
-							( global ) =>
-								/** @type {number} */ window[ global ],
+						const metric = /** @type {number} */ await page.evaluate(
+							( global ) => window[ global ],
 							value.global
 						);
 						value.results.push( metric );
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -415,14 +415,21 @@ async function benchmarkURL( url, browser, metricsDefinition, params ) {
 		}
 
 		if ( groupedMetrics.serverTiming ) {
-			const serverTimingMetrics = await page.evaluate( () =>
-				performance
-					.getEntries()[ 0 ]
-					.serverTiming.reduce( ( acc, value ) => {
-						acc[ value.name ] = value.duration;
-						return acc;
-					}, {} )
-			);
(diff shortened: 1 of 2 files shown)


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -155,15 +166,114 @@ function getParamsFromOptions( opt ) {
...
+			type: 'aggregate',
+			add: [ 'LCP' ],
+			subtract: [ 'TTFB' ],
+		},
+	};
+
+	/*
+	 * Server-Timing metrics can have any name, so for those a generic definition creator is used.
+	 * These metrics must be prefixed with "ST:", see below.
+	 */
```
westonruter, line 207:
```suggestion
	/**
	 * Server-Timing metrics can have any name, so for those a generic definition creator is used.
	 * These metrics must be prefixed with "ST:", see below.
	 *
	 * @param {string} metric
	 * @return {{type: 'serverTiming', name: string}}
	 */
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -180,55 +290,38 @@ export async function handler( opt ) {
 }
 
 /**
+ * @param {string}  url
  * @param {Browser} browser
+ * @param {Object}  metricsDefinition
```
westonruter, line 295: See note above about typedef

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -242,117 +335,165 @@ async function benchmarkURL( browser, params ) {
...
-			metrics[ key ] = value.results;
+		if ( groupedMetrics.serverTiming ) {
+			const serverTimingMetrics = await page.evaluate( () =>
+				performance
+					.getEntries()[ 0 ]
+					.serverTiming.reduce( ( acc, value ) => {
+						acc[ value.name ] = value.duration;
+						return acc;
+					}, {} )
+			);
```
westonruter, line 412: This should be more robust to ensure that the performance entry is indeed of the expected type:

```suggestion
			const serverTimingMetrics = await page.evaluate( () => {
				const entry = performance.getEntries().find(
					( entry ) => entry instanceof PerformanceNavigationTiming
				);
				if ( entry instanceof PerformanceNavigationTiming ) {
					return entry.serverTiming.reduce(
						( acc, value ) => {
							acc[ value.name ] = value.duration;
							return acc;
						}, {}
					);
				} else {
					return {};
				}
			} );
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -242,117 +335,165 @@ async function benchmarkURL( browser, params ) {
...
+	/*
+	 * Include only all the metrics which were requested by the command parameter.
+	 * While the metrics definition is already limited by the parameter for efficiency,
+	 * this logic here is needed because dependencies of aggregate metrics may have been calculated but still shouldn't
+	 * be part of the final list.
+	 */
+	const metrics = {};
+	params.metrics.forEach( ( metric ) => {
+		metrics[ metric ] = metricResults[ metric ];
+	} );
```
westonruter, line 496: Nit: Can be done without a closure too...

```suggestion
	for ( const metric of params.metrics ) {
		metrics[ metric ] = metricResults[ metric ];
	}
```

cli/commands/benchmark-web-vitals.mjs
westonruter: Or even this could return a `MetricsDefinitionEntry` type.

westonruter (changes requested):

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -180,55 +303,38 @@ export async function handler( opt ) {
 }
 
 /**
- * @param {Browser} browser
- * @param {Params}  params
+ * @param {string}                 url
+ * @param {Browser}                browser
+ * @param {MetricsDefinitionEntry} metricsDefinition
+ * @param {Params}                 params
```
westonruter, line 309:
```suggestion
@param {string}                                 url
@param {Browser}                                browser
@param {Object<string, MetricsDefinitionEntry>} metricsDefinition
@param {Params}                                 params
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -155,15 +176,117 @@ function getParamsFromOptions( opt ) {
...
+	 */
+	const getServerTimingDefinition = ( metric ) => {
+		return {
+			type: 'serverTiming',
+			name: metric,
+		};
+	};
+
+	// Set up object with the requested metrics, and store aggregate metrics in a list.
+	const metricsDefinition = {};
```
westonruter, line 229:
```suggestion
	/** @type {Object<string, MetricsDefinitionEntry>} */
	const metricsDefinition = {};
```

Without this, below I get a TypeScript error:

image

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -155,15 +176,117 @@ function getParamsFromOptions( opt ) {
...
+
+	/**
+	 * Server-Timing metrics can have any name, so for those a generic definition creator is used.
+	 * These metrics must be prefixed with "ST:", see below.
+	 *
+	 * @param {string} metric
+	 * @return {MetricsDefinitionEntry} Server-Timing metrics definition entry.
+	 */
+	const getServerTimingDefinition = ( metric ) => {
+		return {
```
westonruter, line 222: I'm getting a TypeScript error here:

image

But I don't understand why, since these `MetricsDefinitionEntry` properties are all nullable.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -155,15 +176,117 @@ function getParamsFromOptions( opt ) {
...
+			type: 'serverTiming',
+			name: metric,
```
westonruter, line 224: This fixes the TypeScript error, but again, I don't see why it should be required:

```suggestion
			name: metric,
			listen: null,
			global: null,
			add: null,
			subtract: null,
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -242,117 +348,170 @@ async function benchmarkURL( browser, params ) {
...
+						await page.click( 'body', {
+							offset: { x: -500, y: -500 },
+						} );
+						// Get the metric value from the global.
+						/** @type {number} */
+						const metric = await page.evaluate(
+							( global ) =>
+								/** @type {number} */ window[ global ],
+							value.global
+						);
```
westonruter, line 408: This fixes another TypeScript error:

```suggestion
						const metric = /** @type {number} */ await page.evaluate(
							( global ) => window[ global ],
							value.global
						);
```

image

cli/commands/benchmark-web-vitals.mjs
felixarntz: Why the different comment format? I followed the one already present above. Is that also incorrect?

westonruter: Different format because it's a different value, right? Here it's passing the object mapping keys to the individual `MetricsDefinitionEntry` objects. Without this change, TypeScript was alerting me of a type error.

image

So the `@param` type here should match whatever is returned by `getMetricsDefinition()`, right?

felixarntz: @westonruter I was referring to you removing the ` * ` from each comment line. The `@param {Object<string, MetricsDefinitionEntry>} metricsDefinition` makes sense to me, but why remove the ` * `?

westonruter: Oh sorry, that was a mistake.

felixarntz: Updated in https://github.com/GoogleChromeLabs/wpp-research/pull/78/commits/6ebf8019c4fb7a1ce508c46cae2311cdc1d028bc

westonruter (approved):

swissspidy (approved):

PR #79, opened 2023-10-26
Log progress when benchmarking multiple URLs

felixarntz:
This PR enhances the `benchmark-server-timing` and `benchmark-web-vitals` commands in two ways:
1. In case a file is provided (i.e. most likely multiple URLs are being benchmarked), the commands will now log progress.
2. Errors in the process are now caught, which is particularly relevant for `benchmark-web-vitals`. Puppeteer errors such as navigation timeouts could otherwise lead to the entire process getting stuck.

While logging isn't really needed for benchmarking a single URL, when benchmarking several URLs the lack of it is a poor experience, as you have no idea on the progress. The second point is particularly relevant for that, as without that it could be possible for that progress to even get completely stuck, and you would never find out about it.

### DIFF
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -86,10 +86,10 @@ export async function handler( opt ) {
 	const results = [];
 
 	// Log progress only under certain conditions (multiple URLs to benchmark).
-	const logProgress = shouldLogURLProgress( opt );
+	const logURLProgress = shouldLogURLProgress( opt );
 
 	for await ( const url of getURLs( opt ) ) {
-		if ( logProgress ) {
+		if ( logURLProgress ) {
 			logPartial( `Benchmarking URL ${ url }...` );
 		}
 
@@ -101,7 +101,7 @@ export async function handler( opt ) {
 					amount,
 				} );
 			results.push( [ url, completeRequests, responseTimes, metrics ] );
-			if ( logProgress ) {
+			if ( logURLProgress ) {
 				log( formats.success( 'Success.' ) );
 			}
 		} catch ( err ) {
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -25,7 +25,7 @@ import round from 'lodash-es/round.js';
 /**
  * Internal dependencies
  */
-import { getURLs } from '../lib/cli/args.mjs';
+import { getURLs, logURLProgress } from '../lib/cli/args.mjs';
 import {
 	log,
 	formats,
@@ -83,16 +83,28 @@ export async function handler( opt ) {
 	const { concurrency: connections, number: amount } = opt;
 	const results = [];
 
-	for await ( const url of getURLs( opt ) ) {
-		const { completeRequests, responseTimes, metrics } = await benchmarkURL(
-			{
-				url,
-				connections,
-				amount,
-			}
-		);
+	// Log progress only under certain conditions (multiple URLs to benchmark).
+	const logProgress = logURLProgress( opt );
 
-		results.push( [ url, completeRequests, responseTimes, metrics ] );
+	for await ( const url of getURLs( opt ) ) {
+		if ( logProgress ) {
+			log( `Benchmarking URL ${ url }...` );
+		}
+
+		try {
+			const { completeRequests, responseTimes, metrics } =
+				await benchmarkURL( {
+					url,
+					connections,
+					amount,
+				} );
+			results.push( [ url, completeRequests, responseTimes, metrics ] );
+			if ( logProgress ) {
+				log( formats.success( 'Success.' ) );
+			}
+		} catch ( err ) {
+			log( formats.error( `Error: ${ err.message }.` ) );
+		}
 	}
 
 	if ( results.length === 0 ) {
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -28,6 +28,8 @@ import round from 'lodash-es/round.js';
 import { getURLs, shouldLogURLProgress } from '../lib/cli/args.mjs';
 import {
 	log,
+	logPartial,
+	output,
 	formats,
 	table,
 	isValidTableFormat,
@@ -88,7 +90,7 @@ export async function handler( opt ) {
 
 	for await ( const url of getURLs( opt ) ) {
 		if ( logProgress ) {
-			log( `Benchmarking URL ${ url }...` );
+			logPartial( `Benchmarking URL ${ url }...` );
 		}
 
 		try {
@@ -264,5 +266,5 @@ function outputResults( opt, results ) {
 		tableData.push( tableRow );
 	}
 
-	log( table( headings, tableData, opt.output, true ) );
+	output( table( headings, tableData, opt.output, true ) );
 }
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -25,7 +25,7 @@ import round from 'lodash-es/round.js';
 /**
  * Internal dependencies
  */
-import { getURLs, logURLProgress } from '../lib/cli/args.mjs';
+import { getURLs, shouldLogURLProgress } from '../lib/cli/args.mjs';
 import {
 	log,
 	formats,
@@ -84,7 +84,7 @@ export async function handler( opt ) {
 	const results = [];
 
 	// Log progress only under certain conditions (multiple URLs to benchmark).
-	const logProgress = logURLProgress( opt );
+	const logProgress = shouldLogURLProgress( opt );
 
 	for await ( const url of getURLs( opt ) ) {
 		if ( logProgress ) {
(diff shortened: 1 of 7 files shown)


cli/lib/cli/args.mjs
```diff
@@ -60,3 +60,7 @@ export async function* getURLs( opt ) {
 		}
 	}
 }
+
+export function logURLProgress( opt ) {
+	return ! opt.url && !! opt.file;
```
westonruter, line 65: It could be that there is only a single URL in the file. Would that be a problem?

cli/lib/cli/args.mjs
```diff
@@ -60,3 +60,7 @@ export async function* getURLs( opt ) {
 		}
 	}
 }
+
+export function logURLProgress( opt ) {
```
westonruter, line 64: Maybe this should be renamed to reflect its boolean nature, like `shouldLogProgress`. Otherwise the name sounds like _it_ is responsible for the logging.

cli/commands/benchmark-server-timing.mjs
```diff
@@ -83,16 +83,28 @@ export async function handler( opt ) {
...
 	for await ( const url of getURLs( opt ) ) {
-		const { completeRequests, responseTimes, metrics } = await benchmarkURL(
-			{
-				url,
-				connections,
-				amount,
-			}
-		);
+		if ( logProgress ) {
+			log( `Benchmarking URL ${ url }...` );
```
westonruter, line 91: Instead of logging out to STDOUT, can this instead go to STDERR? In that way, if you pipe the output into a file, it will by default not include the logging information. This can be achieved by using `console.warn()` instead of `console.log()`. Or it could use `process.stderr.write()`.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -161,13 +161,28 @@ export async function handler( opt ) {
...
+	// Log progress only under certain conditions (multiple URLs to benchmark).
+	const logProgress = logURLProgress( opt );
+
 	for await ( const url of getURLs( opt ) ) {
-		const { completeRequests, metrics } = await benchmarkURL(
-			browser,
-			params
-		);
+		if ( logProgress ) {
+			log( `Benchmarking URL ${ url }...` );
```
westonruter, line 169: Using `process.stderr.write` won't automatically add a newline, so if doing this:

```suggestion
			process.stderr.write( `Benchmarking URL ${ url }...` );
```

Then down below the success/error can be displayed on the same line.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -161,13 +161,28 @@ export async function handler( opt ) {
...
+		// Catch Puppeteer errors to prevent the process from getting stuck.
+		try {
+			const { completeRequests, metrics } = await benchmarkURL(
+				url,
+				browser,
+				params
+			);
+			results.push( [ url, completeRequests, metrics ] );
+			if ( logProgress ) {
+				log( formats.success( 'Success.' ) );
```
westonruter, line 181:
```suggestion
				process.stderr.write( formats.success( `Success\n` ) );
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -161,13 +161,28 @@ export async function handler( opt ) {
...
+			}
+		} catch ( err ) {
+			log( formats.error( `Error: ${ err.message }.` ) );
```
westonruter, line 184:
```suggestion
			process.stderr.write( formats.error( `Error: ${ err.message }.\n` ) );
```

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -161,13 +161,28 @@ export async function handler( opt ) {
...
-			params
-		);
+		if ( logProgress ) {
+			log( `Benchmarking URL ${ url }...` );
+		}
 
-		results.push( [ url, completeRequests, metrics ] );
+		// Catch Puppeteer errors to prevent the process from getting stuck.
+		try {
+			const { completeRequests, metrics } = await benchmarkURL(
```
westonruter, line 174: What about progress for benchmarking an individual URL? If I have 100 requests for each URL, I'd love to be able to see progress for how far along it is when analyzing each URL.

westonruter (changes requested):

cli/lib/cli/args.mjs
felixarntz: I considered that, but it's not really a big deal. Logging too much isn't an issue, but I think it's reasonable to expect that when a file is provided you want to benchmark multiple URLs. Otherwise you could just provide the URL.

Checking whether there's actually one URL in the file would be overly complicated and involve changing how the file contents are parsed, which right now relies on `yield` for more efficient handling. So I think it's not worth going into that.

cli/commands/benchmark-server-timing.mjs
felixarntz: What would be the benefit of that? You'd still end up with the "Success." in the file. Or even if we also moved that to STDOUT, the actual errors would end up in the file.

There is probably room to improve this, but if we want to do so, I think we should enhance the existing `log` and `formats` functions available rather than hard-coding in a specific command.

cli/commands/benchmark-web-vitals.mjs
felixarntz: It looks like this would also be the case for `process.stdout.write`, this is related to the other comments above. But I definitely like the idea of keeping this in one line.

I'm not sure what is the tradeoffs between using `console.*` functions vs `process.stdout/stderr` in this context. We should probably discuss that and then decide how to proceed by updating the `log` and `formats` functions accordingly.

cli/commands/benchmark-web-vitals.mjs
felixarntz: Makes sense to add this. I would suggest we only do so if more than 1 request is being made, as otherwise there wouldn't be any value.

cli/commands/benchmark-server-timing.mjs
westonruter: The benefit is that the actual output wouldn't get intermixed with the logging. For example, you could do:

```bash
npm run --silent research benchmark-server-timing -- --file=urls.txt --output=csv > results.csv
```

When sending logging information to STDERR, then the CSV file will be pristine and the logging information will only be displayed in the console. Same goes for "Success" below.

And STDERR is not just for errors. It's also for diagnostics. cf. https://stackoverflow.com/a/4919110/93579

cli/lib/cli/args.mjs
westonruter: OK, sounds good. Still, it could be useful even when checking a single `--url`. In that case, what about instead of conditionally logging based on whether `--file` is provided, that there be a new `--verbose` option?

cli/commands/benchmark-server-timing.mjs
felixarntz: Makes sense. So I guess we could use either `console.warn()` or `process.stderr.write()`.

felixarntz:
@westonruter I've only addressed one of your feedback points so far, since I think the others deserve a bit more conversation on how to best proceed. For example regarding https://github.com/GoogleChromeLabs/wpp-research/pull/79#discussion_r1374891590 and https://github.com/GoogleChromeLabs/wpp-research/pull/79#discussion_r1373933060.

Let's figure that out here. Once we've made a decision, we can update / expand the log functions in the helper file accordingly, and then the commands' use of them.

felixarntz:
@westonruter I've thought about the logging and output separation a bit more based on your suggestions. How about we implement the following functions:

```js
// Responsible for the actual command output.
export const output = ( text ) => {
	process.stdout.write( `${ text }\n` );
};

// Responsible for (debug) logging.
export const log = ( text ) => {
	process.stderr.write( `${ text }\n` );
};

// Responsible for (debug) logging without "terminating" the line.
export const logPartial = ( text ) => {
	process.stderr.write( text );
};
```

We could then update the usage to rely on the relevant functions. Most importantly, we'd need to change the actual output functions of the command to use `output()`. Let me know what you think - also happy to discuss alternative function names.

westonruter:
@felixarntz Those functions look good to me.

cli/lib/cli/args.mjs
felixarntz: I would prefer not to introduce that for now, unless needed. Why would the existing logging be useful for a single URL? I agree logging would be useful if there are multiple runs, and that shouldn't be coupled to how many URLs are being benchmarked. But the information of "Benchmarking {url}" doesn't provide any value IMO if it's only a single URL.

westonruter: Right, this only makes sense when showing progress for multiple runs of a single URL.

felixarntz:
@westonruter I have added the new more granular logging functions, updated the commands to use them correctly, and I've also added support for logging multiple `benchmark-web-vitals` iterations as requested in https://github.com/GoogleChromeLabs/wpp-research/pull/79#discussion_r1373935911. Can you please take another look?

swissspidy (approved):

cli/lib/cli/logger.mjs
```diff
@@ -23,7 +23,20 @@ import chalk from 'chalk';
...
+
+// Responsible for (debug) logging.
+export const log = ( text ) => {
+	process.stderr.write( `${ text }\n` );
+};
+
+// Responsible for (debug) logging without "terminating" the line.
+export const logPartial = ( text ) => {
+	process.stderr.write( text );
+};
```
westonruter, line 39: Nit: The comments could be formatted as jsdoc so that they show up when inspecting.

westonruter (approved):
Love the logging of the iterations!

PR #80, opened 2023-10-31
Add support for reporting variance in benchmarks

joemcgill:
This adds a new CLI arg to both the benchmark CLI commands that will report the standard deviation (SD) and interquartile range (IQR) for a set of results by passing either `-v` or `--show-variance` as an argument when running a benchmark command.

Example:

```
npm run research benchmark-server-timing -- -u http://localhost:8889/ -p -n 100 -v

╔═════════════════════╤════════════════════════╗
║ URL                 │ http://localhost:8889/ ║
╟─────────────────────┼────────────────────────╢
║ Success Rate        │ 100%                   ║
╟─────────────────────┼────────────────────────╢
║ Response Time (p10) │ 131.22                 ║
╟─────────────────────┼────────────────────────╢
║ Response Time (p25) │ 136.7                  ║
╟─────────────────────┼────────────────────────╢
║ Response Time (p50) │ 142.71                 ║
╟─────────────────────┼────────────────────────╢
║ Response Time (p75) │ 147.52                 ║
╟─────────────────────┼────────────────────────╢
║ Response Time (p90) │ 159.02                 ║
╟─────────────────────┼────────────────────────╢
║ Response Time (SD)  │ 26.7                   ║
╟─────────────────────┼────────────────────────╢
║ Response Time (IQR) │ 10.82                  ║
╚═════════════════════╧════════════════════════╝
```

### DIFF
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -33,7 +33,7 @@ import {
 	isValidTableFormat,
 	OUTPUT_FORMAT_TABLE,
 } from '../lib/cli/logger.mjs';
-import { calcPercentile } from '../lib/util/math.mjs';
+import { calcPercentile, calcStandardDeviation } from '../lib/util/math.mjs';
 import {
 	KEY_PERCENTILES,
 	MEDIAN_PERCENTILES,
@@ -67,6 +67,10 @@ export const options = [
 		argname: '-p, --show-percentiles',
 		description:
 			'Whether to show more granular percentiles instead of only the median',
+	},
+	{
+		argname: '-v, --show-variance',
+		description: 'Whether to show standard deviation and IQR',
 	},
 ];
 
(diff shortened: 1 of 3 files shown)


joemcgill:
Playing with this idea in our compare-wp-performance repo, you can see how it could be useful when interpreting a set of data ([example](https://github.com/joemcgill/compare-wp-performance/actions/runs/6712400590)). For server timing, it seems like there is very low variance in the results population, so it can be inferred that the improvement percentage (which is larger than the IQR) is likely significant and not just the result of variability. Similarly, this run shows an improvement at LCP-TTFB (p50) of -1.75ms, but with both results having SDs of over 6ms and IQRs of nearly 10ms, this could be the result of variance.

cli/commands/benchmark-server-timing.mjs
```diff
@@ -206,10 +210,20 @@ function outputResults( opt, results ) {
...
 		Object.keys( allMetricNames ).forEach( ( metricName ) => {
 			percentiles.forEach( ( percentile ) => {
 				headings.push( `${ metricName } (p${ percentile })` );
 			} );
+			if ( opt.showVariance ) {
+				headings.push( `${ metricName } (SD)` );
+				headings.push( `${ metricName } (IQR)` );
+			}
 		} );
 	} else {
```
felixarntz, line 228: Can you add this to the `else` right below here as well? Otherwise, the `showVariance` would only work when also enabling `showPercentiles`, which I don't think should be needed.

It may actually lead to an error right now because this is missing. Have you tested this without `showPercentiles` but with `showVariance`?

felixarntz (changes requested):
@joemcgill This looks great, there's only one oversight which at the moment causes an error under certain conditions in `benchmark-server-timing`.

swissspidy (approved):
I would find median absolute deviation helpful in addition to standard deviation, as the latter is skewed by single outliers because it's using the mean.

cli/commands/benchmark-server-timing.mjs
joemcgill: Good catch, I hadn't tested this without percentiles, it seems. Should be all good now.

joemcgill:
@swissspidy I've added MAD calculations in [5ae442c](https://github.com/GoogleChromeLabs/wpp-research/pull/80/commits/5ae442c93260f4452fd8c77c4f2a6f816d1b8e12)

felixarntz (approved):
Thanks for the updates @joemcgill!

PR #84, opened 2024-02-01
Add query for the counts of embed blocks on root and non-root pages

westonruter:
For the oEmbed Optimizer project (https://github.com/WordPress/performance/pull/886), I was curious if the count of embeds on non-root pages would be significantly higher. Well, it is _more_ common but only slightly:

is_root_page | has_embed_block | page_count | percentage_of_total
-- | -- | -- | --
TRUE | FALSE | 9858816 | 51.1
FALSE | FALSE | 9190040 | 47.6
FALSE | TRUE | 150877 | 0.8
TRUE | TRUE | 109338 | 0.6

According to this, root pages have embed blocks **1.09%** of the time (109338/(109338+9858816), and non-root pages have them **1.61%** of the time (150877/(150877+9190040)), so %0.52 more often.

### DIFF
--- a/sql/2024/01/embed-blocks-on-root-and-non-root-pages.sql
+++ b/sql/2024/01/embed-blocks-on-root-and-non-root-pages.sql
@@ -1,0 +1,31 @@
+# HTTP Archive query to get scripts blocking in head, counted by plugin.
+#
+# WPP Research, Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SELECT
+  is_root_page,
+  JSON_EXTRACT(custom_metrics, '$.cms.wordpress.has_embed_block') AS has_embed_block,
+  COUNT(*) AS page_count,
+  ROUND(100.0 * COUNT(*) / SUM(COUNT(*)) OVER (), 1) AS percentage_of_total
+FROM
+  `httparchive.all.pages`,
+  UNNEST(technologies) AS technology
+WHERE
+  date = CAST('2023-12-01' AS DATE)
+  AND technology.technology = 'WordPress'
+GROUP BY
+  has_embed_block,
+  is_root_page
+ORDER BY percentage_of_total DESC
--- a/sql/2024/01/embed-blocks-on-root-and-non-root-pages.sql
+++ b/sql/2024/01/embed-blocks-on-root-and-non-root-pages.sql
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


westonruter:
Naturally this query doesn't factor in all embeds coming from the classic editor. But there is no uniform way to identify them. I wonder what the ratio is for WP pages using blocks versus those which aren't. That could be a good signal to weight these numbers, or we could try excluding all pages that have no blocks on them at all.

westonruter:
@adamsilverstein I tried adding this `WHERE` condition to restrict the results to just those with block themes:

```sql
AND JSON_EXTRACT(custom_metrics, '$.cms.wordpress.block_theme') = "true"
```

is_root_page | has_embed_block | page_count | percentage_of_total
-- | -- | -- | --
TRUE | FALSE | 113777 | 49.9
FALSE | FALSE | 102461 | 44.9
TRUE | TRUE | 6220 | 2.7
FALSE | TRUE | 5560 | 2.4

So here, **5.18%** (6220 / (6220 + 113777)) of root pages have embed blocks, and **5.15%** (5560 / (5560 + 102461)) of non-root pages have embed blocks.

Assuming that sites with classic themes and block themes are equally likely to include embeds and that sites with block themes never use the Classic block, then it seems the ~1% figure is significantly under-counting the number of embeds that actually exist on WordPress sites. Namely, it seems that perhaps only a fifth to a third (21.04% = 1.09% / 5.18% to 31.26% = 1.61% / 5.15%) of embeds on WordPress sites are actually using the embed block. The rest were embedded in the classic way.

adamsilverstein:
> Namely, it seems that perhaps only a fifth to a third (21.04% = 1.09% / 5.18% to 31.26% = 1.61% / 5.15%) of embeds on WordPress sites are actually using the embed block. The rest were embedded in the classic way.

thanks for doing this additional research. That matches my intuition about the frequency of embeds and probably reflects the large number of existing sites with embeds that were created using the classic editor. 

Use of the embed block is more tied to the editor than the theme though, right? ie. If I embed a url into a classic theme using the block editor, it still winds up being an oembed block.

adamsilverstein (approved):

swissspidy (approved):

westonruter:
@adamsilverstein: 

> Use of the embed block is more tied to the editor than the theme though, right? ie. If I embed a url into a classic theme using the block editor, it still winds up being an oembed block.

I don't think that is the case. I just created a Classic block and dropped in a YouTube URL. When saving the Classic block and going back to the editor, I see:

> image

The `post_content` is:

```html
Check out this video:

https://www.youtube.com/watch?v=3KtWfp0UopM&amp;t=4s
```

And on the frontend it is rendered as:

```html
<p>Check out this video:</p>
<p><iframe loading="lazy" title="Google — 25 Years in Search: The Most Searched" width="500" height="281" src="[https://www.youtube.com/embed/3KtWfp0UopM?start=4&#038;feature=oembed](https://www.youtube.com/embed/3KtWfp0UopM?start=4&feature=oembed)" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" allowfullscreen></iframe></p>
```

So yeah, it isn't converted to blocks automatically.

westonruter:
@adamsilverstein Oh, I misunderstood what you wrote. Yes, adding an embed in the block editor for a classic theme will indeed cause the embed to be rendered in an embed block. 

I was trying to use the block theme as a proxy for the likelihood that a site would have fully adopted blocks, and that they would be even less likely to be using the classic _editor_ since they aren't using a classic theme. But yeah, even when only querying block themes, this is also likely an undercount since it is possible (but I think unlikely) that there would be many classic blocks on a block theme.

PR #162, opened 2024-10-03
Implement SQL query to compare impact of using auto-sizes plugin for `sizes=auto` functionality

felixarntz:
This PR's query takes the approach from #108 and uses it in a comparison over time, specifically focused on determining the impact of the `sizes="auto"` addition from the [Enhanced Responsive Images plugin (slug: `auto-sizes`)](https://wordpress.org/plugins/auto-sizes/). The relevant feature was [recently committed to WordPress Core](https://core.trac.wordpress.org/changeset/59008).

It compares sites that were using the plugin in 2024-06 with the same intersection of sites _not_ using the plugin in 2024-05. It then also considers only images with the same URL on the same site, to get as close to a consistent dataset as possible.

The timeframe was chosen because it needs to be before 2024-07, as in July version 1.1.0 of the plugin introduced additional functionality which we don't want to influence the result. Additionally, the choice of comparing between adjacent months was made so that there's the least possible room for changes on each site that could skew the data.

## Query results

Since the raw query results are so large, please see [this spreadsheet](https://docs.google.com/spreadsheets/d/1ea2AKpGHC1cajkYHt9fiNyrQsIYXRwzn5GK8P32UcD4/edit).

The spreadsheet also contains a summary of the diff in another tab, as well as the results of a similar query that gets the LCP passing rate of these sites before and after, as another reference point. For the CrUX data, it uses the approach from https://colab.research.google.com/drive/1TmbJhSBuM-gTHA_E_W96pSjraQkFYpyi#scrollTo=XoNNe9GNW2hm.

### DIFF
--- a/sql/2024/10/auto-sizes-impact-before-after.sql
+++ b/sql/2024/10/auto-sizes-impact-before-after.sql
@@ -1,0 +1,188 @@
+# HTTP Archive query to compare the image sizes attribute impact of using the `auto-sizes` plugin.
+#
+# WPP Research, Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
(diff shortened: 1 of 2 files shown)


joemcgill:
Thanks! I think this is a nice way of seeing the impact.

It's worth noting that this query will include images that do not get auto-sizes applied (e.g. non lazy-loaded images), even when the plugin is added so the impact on affected images might be even greater. Would that require a new custom metric, like `image.hasAutoSizes`? If so, should we try adding a new custom metric for that prior to the WP 6.7 release?

It's also nice to see a positive correlation between adding this plugin and improved LCP passing rates. There are other possible factors that could have affected the LCP passing rates on those URLs, but this certainly is a positive sign.

westonruter (approved):

joemcgill (approved):

PR #166, opened 2024-11-26
Add device emulation to `benchmark-web-vitals`

westonruter:
This is a follow-up to #164.

Being able to specify the window dimensions alone is not always sufficient to test a page. In particular, there is also the device scale factor which is often higher on mobile devices. More importantly, if the user agent is not indicating a mobile device then a site may return a page formatted for desktop. This PR introduces a `--emulate-device` (`-e`) parameter to specify one of the [known devices](https://pptr.dev/api/puppeteer.knowndevices) in Puppeteer. The `--window-viewport` arg may still be supplied, and it will override the `width` and `height` of the emulated device. If no device is being emulated, then the default viewport remains 960x700.

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -19,13 +19,17 @@
 /**
  * External dependencies
  */
-import puppeteer, { Browser, PredefinedNetworkConditions } from 'puppeteer';
+import puppeteer, { Browser, PredefinedNetworkConditions, KnownDevices } from 'puppeteer';
 import round from 'lodash-es/round.js';
 
 /* eslint-disable jsdoc/valid-types */
 /** @typedef {import("puppeteer").NetworkConditions} NetworkConditions */
 /** @typedef {keyof typeof PredefinedNetworkConditions} NetworkConditionName */
+/** @typedef {import("puppeteer").Device} Device */
+/** @typedef {keyof typeof KnownDevices} KnownDeviceName */
+
 /* eslint-enable jsdoc/valid-types */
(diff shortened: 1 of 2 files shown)


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -449,12 +468,15 @@ async function benchmarkURL(
 			await page.emulateNetworkConditions( params.networkConditions );
 		}
 
-		await page.setViewport( params.windowViewport );
-		await page
-			.mainFrame()
-			.waitForFunction(
-				`window.innerWidth === ${ params.windowViewport.width } && window.innerHeight === ${ params.windowViewport.height }`
-			);
```
westonruter, line 457: I removed this code because it with the `deviceScaleFactor` coming from device emulation, I found the `innerWidth` and `innerHeight` to be the viewport's dimensions multiplied by the device scale factor. While I could implement that adjustment, I didn't find that this code was actually necessary.

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -338,7 +357,7 @@ export async function handler( opt ) {
 	const params = getParamsFromOptions( opt );
 	const results = [];
 
-	const browser = await puppeteer.launch( { headless: 'new' } );
+	const browser = await puppeteer.launch( { headless: true } );
```
westonruter, line 360: This change was caught by TypeScript since Puppeteer the new headless mode is now the default.

felixarntz (approved):
@westonruter Thanks, LGTM!

felixarntz:
@westonruter Looks like there are some lint failures: https://github.com/GoogleChromeLabs/wpp-research/actions/runs/12038260972/job/33563535827?pr=166

PR #167, opened 2024-12-16
Implement SQL query to assess impact of WordPress 6.7 image `sizes=auto` functionality

felixarntz:
This PR reuses most of the query from #162, but instead of comparing sites with and without the `auto-sizes` WordPress plugin, it compares sites before and after updating to WordPress 6.7, which includes the `sizes=auto` functionality.

## Query results

Since the raw query results are so large, please see [this spreadsheet](https://docs.google.com/spreadsheets/d/1OtONUyfsjn66LKbbHmG5cyNgG0nePUdXvEEE6saxYVY/edit).

While relative improvements are smaller than in the `auto-sizes` plugin specific query, this was expected given the much larger number of sites. The impact is definitely considerable, especially on desktop. For mobile, the impact is small (but still net-positive), which makes sense because mobile viewports are narrow and thus the assumption that an image spans (almost) 100% of the viewport is more likely to be accurate than on desktop, where websites usually use a multi-column layout in some capacity.

### DIFF
--- a/sql/2024/12/auto-sizes-wp67-impact-before-after.sql
+++ b/sql/2024/12/auto-sizes-wp67-impact-before-after.sql
@@ -1,0 +1,162 @@
+# HTTP Archive query to compare the image sizes attribute impact of using WordPress 6.7.
+#
+# WPP Research, Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


adamsilverstein (approved):
Great!

joemcgill (approved):
👍🏻

PR #174, opened 2025-01-08
Update existing queries to use latest `httparchive.crawl.*` dataset

felixarntz:
2 months ago, HTTP Archive introduced a new dataset `httparchive.crawl.*`, which going forward will be the recommended way to query HTTP Archive. [See the announcement.](https://discuss.httparchive.org/t/new-release-httparchive-crawl-dataset/2814)

As of `2024-12-01`, the older datasets (`httparchive.all.*`, and the older monthly tables) are no longer being populated. Furthermore the old datasets will soon be deleted entirely, as mentioned in the announcement.

Since the WPP Research repository is intended as a repository of queries to use for reference, it makes sense to update the queries in the repository to use the new table structure. While we probably don't intend to use most of the existing queries again as is, them serving as a reference is a good reason to make sure they _can_ actually serve as a reference - which as of now is no longer the case. For queries that were already using `httparchive.all.*` this is a very small lift, while for the older queries using the monthly tables it's a bit more effort.

This PR migrates all queries to use the `httparchive.crawl.*` dataset, with the minimal changes needed. Of course some older queries could in principle be optimized, but this is out of scope for the PR. The goal is to simply migrate to the new dataset while maintaining the same query outputs and with minimal changes.

Only a few queries are not updated (yet) as part of this PR because after careful consideration this was considered too much effort:
* [% of WordPress sites that use any web fonts](https://github.com/GoogleChromeLabs/wpp-research/blob/main/sql/2023/01/web-fonts-usage.sql) (see #34)
* [Distribution of number of web fonts used per site](https://github.com/GoogleChromeLabs/wpp-research/blob/main/sql/2023/01/web-fonts-count-distribution.sql) (see #34)
* [% of WordPress sites that use various font-display strategy for any web fonts](https://github.com/GoogleChromeLabs/wpp-research/blob/main/sql/2023/01/font-display-strategy-usage.sql) (see #34)

I have rerun all the queries with the proposed updates and verified that they produce the same results (+/- <0.1% different values due to minor differences in aggregation of the tables themselves).

For another reference, see the educational [WordPress and performance research in HTTP Archive Colab](https://colab.research.google.com/drive/1TmbJhSBuM-gTHA_E_W96pSjraQkFYpyi) which I also updated the other day to use the new `httparchive.crawl.*` dataset.

Last but not least, there is also [documentation about the migration](https://har.fyi/guides/migrating-to-crawl-dataset/).

### DIFF
--- a/sql/2022/11/performance-lab-version-distribution.sql
+++ b/sql/2022/11/performance-lab-version-distribution.sql
@@ -15,26 +15,18 @@
 # limitations under the License.
 
 SELECT
-  info as version,
-  COUNT(DISTINCT url) AS sites,
-  COUNT(DISTINCT url) / total AS pct_sites
-FROM
-  `httparchive.technologies.2022_10_01_*`
-JOIN (
-  SELECT
-    app,
-    COUNT(DISTINCT url) AS total
-  FROM
-    `httparchive.technologies.2022_10_01_*`
-  GROUP BY
-    app
-)
-USING
-  (app)
-WHERE
-  app = 'Performance Lab'
-GROUP BY
   version,
-  total
+  COUNT(DISTINCT page) AS sites,
+  COUNT(DISTINCT page) / SUM(COUNT(DISTINCT page)) OVER () AS pct_sites
+FROM
+  `httparchive.crawl.pages`,
+  UNNEST(technologies) AS technology,
+  UNNEST(technology.info) AS version
+WHERE
+  date = '2022-10-01'
+  AND is_root_page
+  AND technology.technology = 'Performance Lab'
+GROUP BY
+  version
 ORDER BY
   sites DESC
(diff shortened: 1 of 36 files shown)


felixarntz:
@adamsilverstein @joemcgill @kt-12 @mukeshpanchal27 @swissspidy @westonruter Of course not all of you _need_ to review this PR, but it might be helpful if you could review the updates to the queries that you originally wrote - simply because you have most context on them.

You can trust me that I have run each one of them and verified that it still produces the same results, but a set of familiar eyes on the changes is definitely a good idea.

westonruter (approved):
Thank you for doing this.

mukeshpanchal27 (approved):

swissspidy (approved):
Thanks! Really helpful to see the difference between the datasets across all these queries. Makes it easier to grasp the full change.

joemcgill (approved):
Nice! Thanks for the update, @felixarntz. This looks good to me.

PR #177, opened 2025-03-21
Fix importing web-vitals module from unpkg

westonruter:
See https://github.com/GoogleChrome/web-vitals/pull/600

I also noticed that web-vitals was still using v3 when the current version is v4 (as is reflected in `package.json`, updated in #173). So this updates that version as well.


adamsilverstein (approved):
Nice sleuthing!

PR #180, opened 2025-04-08
Add broadband network conditions emulation to benchmark-web-vitals

westonruter:
This adds a new `broadband` option for emulating network conditions in `benchmark-web-vitals` to correspond with the configuration used with Lighthouse in PageSpeed Insights.


swissspidy (approved):

swissspidy:
@westonruter should we now use that one in https://github.com/swissspidy/compare-wp-performance/?

westonruter:
@swissspidy Yes, I presume so. It should speed up the tests.

PR #181, opened 2025-04-22
Add query for slider/carousel usage

swissspidy:
In the 2025-03-01 dataset, ~54% of all WordPress sites use one of these slider/carousel libraries.

The list of plugins/libraries is based on some manual research and is not meant to be exhaustive.

This is relevant for things like https://core.trac.wordpress.org/ticket/63251

I am sharing these numbers in an upcoming blog post and need to store the query somewhere, so here I am :)

### DIFF
--- a/sql/2025/04/slider-usage.sql
+++ b/sql/2025/04/slider-usage.sql
@@ -1,0 +1,72 @@
+  # HTTP Archive query to get % of WordPress sites using a slider/carousel library.
+  #
+  # WPP Research, Copyright 2025 Google LLC
+  #
+  # Licensed under the Apache License, Version 2.0 (the "License");
+  # you may not use this file except in compliance with the License.
+  # You may obtain a copy of the License at
+  #
+  #     https://www.apache.org/licenses/LICENSE-2.0
+  #
+  # Unless required by applicable law or agreed to in writing, software
+  # distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


sql/2025/04/slider-usage.sql
```diff
@@ -0,0 +1,72 @@
...
+        'Master Slider',
+        'MetaSlider',
+        'OWL Carousel',
+        'Slick',
+        'Slider Revolution',
+        'Smart Slider 3',
+        'Swiper'
+      )
+    )
+  );
```
felixarntz, line 56: This is pretty much the same as `HAS_TECHNOLOGY`, except that it looks for one of multiple technologies.

You could get rid of the two functions in favor of a tiny change to make `HAS_TECHNOLOGY` expect an array of strings instead of a single string. Then you can pass all the technologies to the function.

You could keep `HAS_SLIDER()` if you think it has value but have it call `HAS_TECHNOLOGIES`.

Not a big deal, but avoids duplicate implementation of almost the same piece of query.

felixarntz (approved):
@swissspidy Looks good, just an optional recommendation.

sql/2025/04/slider-usage.sql
swissspidy: Yeah I thought about that but figured it doesn't really matter for a one off query.

PR #184, opened 2025-05-06
Implement HTTP Archive query to assess usage of `Featured Image` in block themes

### LINKED ISSUE #183: Query to assess usage of `Featured Image` in block themes
As part of the [Improving the calculation of image sizes attributes](https://github.com/WordPress/performance/issues/760) project, it would be helpful to understand how widely the `core/post-featured-image` block is used.

Create an HTTP Archive query to determine the **percentage of WordPress sites that include a `figure.wp-block-post-featured-image` element in their markup**

mukeshpanchal27:
Fixes #183 

This query identifies sites using a `Featured Image` in block themes by checking for the `<figure class="wp-block-post-featured-image">` tag which is present in all such sites, as controlled by WordPress core.

## Query results

Row | client | urls | block_theme | total | pct_block_theme | pct_total
-- | -- | -- | -- | -- | -- | --
1 | desktop | 5806 | 99881 | 4573341 | 0.0581291737167229 | 0.0012695313994736015
2 | mobile | 6848 | 127599 | 5595207 | 0.053668132195393382 | 0.0012239046741255506

Based on April 2025 dataset

### DIFF
--- a/sql/2025/05/featured-image-block-usage.sql
+++ b/sql/2025/05/featured-image-block-usage.sql
@@ -34,7 +34,8 @@ CREATE TEMPORARY FUNCTION IS_CMS(technologies ARRAY<STRUCT<technology STRING, ca
 WITH wordpress AS (
   SELECT
     client,
-    page
+    page,
+    BOOL(custom_metrics.cms.wordpress.block_theme) AS has_block_theme
   FROM
     `httparchive.crawl.pages`
   WHERE
@@ -60,7 +61,9 @@ featuredImageBlockDetection AS (
 SELECT
   client,
   COUNT(IF(has_featured_image_block, page, NULL)) AS urls,
+  COUNT(IF(has_block_theme, page, NULL)) AS block_theme,
   COUNT(page) AS total,
+  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(IF(has_block_theme, page, NULL)) AS pct_block_theme,
   COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(page) AS pct_total
 FROM
   wordpress
--- a/sql/2025/05/featured-image-block-usage.sql
+++ b/sql/2025/05/featured-image-block-usage.sql
@@ -15,47 +15,59 @@
 # limitations under the License.
 
 # See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/184
-SELECT
-  client,
-  COUNT(DISTINCT page) AS with_block_theme,
-  total_wp_sites,
-  COUNT(DISTINCT url) / total_wp_sites AS pct_total,
-  # For reference, include number of sites greater than or equal to WP 5.9, since only then block theme support was launched.
-  wp_gte_59
-FROM
-  `httparchive.crawl.pages`,
-  UNNEST(technologies) AS technology,
-  UNNEST(technology.info) AS version
-JOIN
-  `httparchive.crawl.requests`
-USING
-  (date, client, page, is_root_page)
-JOIN (
+
+DECLARE
+  DATE_TO_QUERY DATE DEFAULT '2025-03-01';
+
+CREATE TEMPORARY FUNCTION IS_CMS(technologies ARRAY<STRUCT<technology STRING, categories ARRAY<STRING>, info ARRAY<STRING>>>, cms STRING, version STRING) RETURNS BOOL AS (
+  EXISTS(
+    SELECT * FROM UNNEST(technologies) AS technology, UNNEST(technology.info) AS info
+    WHERE technology.technology = cms
+    AND (
+      version = ""
+      OR ENDS_WITH(version, ".x") AND (STARTS_WITH(info, RTRIM(version, "x")) OR info = RTRIM(version, ".x"))
+      OR info = version
+    )
+  )
+);
+
+WITH wordpress AS (
   SELECT
     client,
-    COUNT(DISTINCT IF (version = '' OR CAST(REGEXP_EXTRACT(version, r'^(\d+\.\d+)') AS FLOAT64) >= 5.9, page, NULL)) AS wp_gte_59,
-    COUNT(DISTINCT page) AS total_wp_sites
+    page
   FROM
-    `httparchive.crawl.pages`,
-    UNNEST(technologies) AS technology,
-    UNNEST(technology.info) AS version
+    `httparchive.crawl.pages`
   WHERE
-    date = '2025-03-01'
-    AND is_root_page
-    AND technology.technology = "WordPress"
-  GROUP BY
-    client )
-USING
-  (client)
-WHERE
-  date = '2025-03-01'
-  AND is_root_page
-  AND technology.technology = "WordPress"
-  AND (version = '' OR CAST(REGEXP_EXTRACT(version, r'^(\d+\.\d+)') AS FLOAT64) >= 5.9)
-  AND response_body LIKE '%<figure class="wp-block-post-featured-image">%'
-GROUP BY
+    date = DATE_TO_QUERY
+    AND IS_CMS(technologies, 'WordPress', '5.9.0')
+    AND is_root_page = TRUE
+),
+
+featuredImageBlockDetection AS (
+  SELECT
+    client,
+    page,
+    REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">') AS has_featured_image_block
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = DATE_TO_QUERY
+    AND is_root_page = TRUE
+    AND is_main_document
+)
+
+SELECT
   client,
-  wp_gte_59,
-  total_wp_sites
+  COUNT(IF(has_featured_image_block, page, NULL)) AS urls,
+  COUNT(page) AS total,
+  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(page) AS pct_total
+FROM
+  wordpress
+JOIN
+  featuredImageBlockDetection
+USING
+  (client, page)
+GROUP BY
+  client
 ORDER BY
   client
--- a/sql/2025/05/featured-image-block-usage.sql
+++ b/sql/2025/05/featured-image-block-usage.sql
@@ -1,0 +1,61 @@
+# HTTP Archive query to get % WordPress sites using a block theme.
+#
+# WPP Research, Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/184
+SELECT
+  client,
+  COUNT(DISTINCT page) AS with_block_theme,
+  total_wp_sites,
+  COUNT(DISTINCT url) / total_wp_sites AS pct_total,
+  # For reference, include number of sites greater than or equal to WP 5.9, since only then block theme support was launched.
+  wp_gte_59
+FROM
+  `httparchive.crawl.pages`,
+  UNNEST(technologies) AS technology,
+  UNNEST(technology.info) AS version
+JOIN
+  `httparchive.crawl.requests`
+USING
+  (date, client, page, is_root_page)
+JOIN (
+  SELECT
+    client,
+    COUNT(DISTINCT IF (version = '' OR CAST(REGEXP_EXTRACT(version, r'^(\d+\.\d+)') AS FLOAT64) >= 5.9, page, NULL)) AS wp_gte_59,
+    COUNT(DISTINCT page) AS total_wp_sites
+  FROM
+    `httparchive.crawl.pages`,
+    UNNEST(technologies) AS technology,
+    UNNEST(technology.info) AS version
+  WHERE
+    date = '2025-03-01'
+    AND is_root_page
+    AND technology.technology = "WordPress"
+  GROUP BY
+    client )
+USING
+  (client)
+WHERE
+  date = '2025-03-01'
+  AND is_root_page
+  AND technology.technology = "WordPress"
+  AND (version = '' OR CAST(REGEXP_EXTRACT(version, r'^(\d+\.\d+)') AS FLOAT64) >= 5.9)
+  AND response_body LIKE '%<figure class="wp-block-post-featured-image">%'
+GROUP BY
+  client,
+  wp_gte_59,
+  total_wp_sites
+ORDER BY
+  client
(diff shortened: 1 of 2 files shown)


mukeshpanchal27:
@swissspidy @felixarntz Could either of you please review this PR and run the query, then share the results in the PR comments?

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,61 @@
...
+  client,
+  COUNT(DISTINCT page) AS with_block_theme,
+  total_wp_sites,
+  COUNT(DISTINCT url) / total_wp_sites AS pct_total,
+  # For reference, include number of sites greater than or equal to WP 5.9, since only then block theme support was launched.
+  wp_gte_59
+FROM
+  `httparchive.crawl.pages`,
+  UNNEST(technologies) AS technology,
+  UNNEST(technology.info) AS version
```
felixarntz, line 28: It's important to be careful with `UNNEST`ing in queries, as it can easily cause duplicated rows. The `UNNEST(technologies)` is definitely fine here because in the `WHERE` clause you limit it to WordPress, but I'm not sure about `UNNEST(technology.info)`.

I would recommend to avoid `UNNEST`ing and instead use a helper function like this one: https://colab.research.google.com/drive/1TmbJhSBuM-gTHA_E_W96pSjraQkFYpyi#scrollTo=BFifI50QnBAs

That way you don't `UNNEST` in the query itself and avoid duplicated rows with certainty.

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,61 @@
...
+JOIN
+  `httparchive.crawl.requests`
+USING
+  (date, client, page, is_root_page)
+JOIN (
```
felixarntz, line 33: As a best practice, `JOIN` statements should always be "concrete", i.e. what kind of join? I assume these should be `INNER JOIN`.

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,61 @@
...
+  FROM
+    `httparchive.crawl.pages`,
+    UNNEST(technologies) AS technology,
+    UNNEST(technology.info) AS version
+  WHERE
+    date = '2025-03-01'
+    AND is_root_page
+    AND technology.technology = "WordPress"
+  GROUP BY
+    client )
```
felixarntz, line 47: This extra `JOIN` is unnecessary and will make the query unnecessarily slow. You should be able to only use one partial query for `httparchive.crawl.pages`, joined with another partial query for `httparchive.crawl.requests`.

Ideally something like this:
* Query for `httparchive.crawl.pages` queries all root pages per client that are using WordPress. For every row, it also includes a boolean whether it's on version 5.9+ or not. Don't do any grouping/counting at this point.
* Query for `httparchive.crawl.requests` queries all root pages (and restrict to only the main document) per client that have the `response_body LIKE '%<figure class="wp-block-post-featured-image">%'`. The columns for this query result should include one extra column that the first query doesn't have, e.g. `has_post_featured_image_block`, which would always be `true` given you filter by it in the `WHERE` clause. Filtering in this inner query will also make the join faster because there will be fewer results.
* Then you do a `LEFT JOIN` between the first query result and the second query result. This will ensure that all results from the `httparchive.crawl.pages` query are included, even if there's no match for them in the second query. The extra column from the second query will then either be `true` or `null` (if there was no match).
* With that joined data, you can then do all the grouping and counting.

One extra recommendation, try using `WITH ...` for the partial queries. Nesting a bunch of inner queries is much harder to read and IMO also harder to get working correctly. By writing the queries separately as `WITH ...` sub queries, you can simply reference the results in the main query.

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,61 @@
...
+  total_wp_sites,
+  COUNT(DISTINCT url) / total_wp_sites AS pct_total,
+  # For reference, include number of sites greater than or equal to WP 5.9, since only then block theme support was launched.
+  wp_gte_59
+FROM
+  `httparchive.crawl.pages`,
+  UNNEST(technologies) AS technology,
+  UNNEST(technology.info) AS version
+JOIN
+  `httparchive.crawl.requests`
```
felixarntz, line 30: One thing this query is missing is limiting the `httparchive.crawl.requests` data to only the `is_main_document` results. `httparchive.crawl.requests` has rows for every single request (e.g. CSS files, JS files, images, etc.), but all we care about here is the HTML request (`is_main_document`).

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,61 @@
+# HTTP Archive query to get % WordPress sites using a block theme.
```
felixarntz, line 1: Can you update this text? As it is not quite correct.

felixarntz (changes requested):
Thanks @mukeshpanchal27! I left a bit of feedback. The query looks like a good start, but there are some unnecessary JOINs and a few other potential issues.

mukeshpanchal27:
@felixarntz thanks for the feedback. I referenced https://github.com/GoogleChromeLabs/wpp-research/blob/main/sql/2023/01/block-theme-usage.sql for the PR. If time permits, could you please help update the query and share the results? I currently don't have access to run it myself.

mukeshpanchal27:
@felixarntz I update query as per https://colab.research.google.com/drive/1TmbJhSBuM-gTHA_E_W96pSjraQkFYpyi#scrollTo=KEQqFjR9FsIF&line=1&uniqifier=1 document.

Could you please take a look. Thanks!

mukeshpanchal27:
Just a quick ping, review when you have a moment. Appreciate it!

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,73 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/184
+
+DECLARE
+  DATE_TO_QUERY DATE DEFAULT '2025-03-01';
```
swissspidy, line 20:
```suggestion
  DATE_TO_QUERY DATE DEFAULT '2025-04-01';
```

Let's use the latest dataset

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,73 @@
...
+
+WITH wordpress AS (
+  SELECT
+    client,
+    page
+  FROM
+    `httparchive.crawl.pages`
+  WHERE
+    date = DATE_TO_QUERY
+    AND IS_CMS(technologies, 'WordPress', '5.9.0')
```
swissspidy, line 42: Note that `5.9.0` would only match exactly this major version. You'd want `5.9.x` to include minor releases as well. However, what we're interested in are all sites on 5.9 _or newer_, not just sites on 5.9 (which would be a very small number). For that we'd need to modify the query slightly to do a version number comparison, or do a few `or is 6.0.x, or is 6.1.x, ...` etc.

But I think we can keep this simple, because most sites would be on newer versions anyway. So let's just ignore the WordPress version number altogether:

```suggestion
    AND IS_CMS(technologies, 'WordPress', '')
```

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,73 @@
...
+    client,
+    page,
+    REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">') AS has_featured_image_block
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = DATE_TO_QUERY
+    AND is_root_page = TRUE
+    AND is_main_document
+)
```
swissspidy, line 57: The query right now looks as featured image block across _all WordPress sites_, which is a low number (~1.2%). What's missing is filtering by block themes.

```suggestion
featuredImageBlockDetection AS (
  SELECT
    client,
    page,
    REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">') AS has_featured_image_block
  FROM
    `httparchive.crawl.requests`
  WHERE
    date = DATE_TO_QUERY
    AND is_root_page = TRUE
    AND is_main_document
    AND response_body LIKE '%<div class="wp-site-blocks">%'
)
```

With this, you'll get more reasonable results:

```

desktop | 19820 | 53448 | 0.37082772040113754 |  
-- | -- | -- | -- | --
2 | mobile | 25700 | 68758 | 0.37377468803630121 |  

```

In other words: 37% of all WordPress sites using a block theme use the featured image block on the homepage.

sql/2025/05/featured-image-block-usage.sql
mukeshpanchal27: Thanks! When I created this PR earlier this month, I believed that dataset wasn't available for querying. But since it's now available, it makes more sense to use the latest version.

sql/2025/05/featured-image-block-usage.sql
mukeshpanchal27: Thanks for sharing the results. Based on the data, we should prioritize updating it to ensure accurate sizes, it's a great opportunity since 37% is a significant portion of sites.

swissspidy (approved):

mukeshpanchal27:
@felixarntz Could you please review it when you have moments. Thanks!

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,74 @@
...
+    AND is_root_page = TRUE
+    AND is_main_document
+    AND response_body LIKE '%<div class="wp-site-blocks">%'
+)
+
+SELECT
+  client,
+  COUNT(IF(has_featured_image_block, page, NULL)) AS urls,
+  COUNT(page) AS total,
+  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(page) AS pct_total
```
felixarntz, line 64: In addition to this metric, it might be helpful to get the % relative to all sites using a block theme? After all, this is what the description of the query says.

I think having % of all WordPress sites is good, so we should keep it too, but it would make sense to capture the % of block theme sites in addition.

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,74 @@
...
+      OR ENDS_WITH(version, ".x") AND (STARTS_WITH(info, RTRIM(version, "x")) OR info = RTRIM(version, ".x"))
+      OR info = version
+    )
+  )
+);
+
+WITH wordpress AS (
+  SELECT
+    client,
+    page
```
felixarntz, line 37: Per my other comment:
```suggestion
  SELECT
    client,
    page,
    BOOL(custom_metrics.cms.wordpress.block_theme) AS has_block_theme
```

sql/2025/05/featured-image-block-usage.sql
felixarntz, line 64: Per my other comment:
```suggestion
SELECT
  client,
  COUNT(IF(has_featured_image_block, page, NULL)) AS urls,
  COUNT(IF(has_block_theme, page, NULL)) AS block_theme,
  COUNT(page) AS total,
  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(IF(has_block_theme, page, NULL)) AS pct_block_theme,
  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(page) AS pct_total
```

felixarntz (commented):
@mukeshpanchal27 This looks good to go technically, but I think it would be good to include another metric, based on what the query's described goal is.

mukeshpanchal27:
@felixarntz Can you please share the result?

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,77 @@
...
+    client,
+    page,
+    REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">') AS has_featured_image_block
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = DATE_TO_QUERY
+    AND is_root_page = TRUE
+    AND is_main_document
+    AND response_body LIKE '%<div class="wp-site-blocks">%'
```
felixarntz, line 58: I missed before that this was included here. While it is the manual way to detect block theme usage, it is unnecessarily slow compared to the `custom_metrics.cms.wordpress.block_theme` custom metric, which has already stored this information as a simple boolean, in the much faster query for `httparchive.crawl.pages`.

In combination with my proposed change from the previous review, this is also unnecessary to keep as it would limit the results to only include block themes due to the `JOIN`. So from that perspective, we should remove this condition.

To make the query faster though, we can use the condition for featured image block presence here instead, in combination with changing the `JOIN` to a `LEFT JOIN`. This way, this sub query only includes the relevant URLs with featured image block, and by `LEFT JOIN` we'll still keep all results from `httparchive.crawl.pages`, even those without featured image. Those without featured image will then have the value `has_featured_image_block: null`, which in an `IF()` check also evaluates to `false`.
```suggestion
    AND REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">')
```

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,77 @@
...
+  client,
+  COUNT(IF(has_featured_image_block, page, NULL)) AS urls,
+  COUNT(IF(has_block_theme, page, NULL)) AS block_theme,
+  COUNT(page) AS total,
+  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(IF(has_block_theme, page, NULL)) AS pct_block_theme,
+  COUNT(IF(has_featured_image_block, page, NULL)) / COUNT(page) AS pct_total
+FROM
+  wordpress
+JOIN
+  featuredImageBlockDetection
```
felixarntz, line 71: See above:
```suggestion
LEFT JOIN
  featuredImageBlockDetection
```

sql/2025/05/featured-image-block-usage.sql
```diff
@@ -0,0 +1,77 @@
...
+    date = DATE_TO_QUERY
+    AND IS_CMS(technologies, 'WordPress', '')
+    AND is_root_page = TRUE
+),
+
+featuredImageBlockDetection AS (
+  SELECT
+    client,
+    page,
+    REGEXP_CONTAINS(response_body, r'<figure class="wp-block-post-featured-image">') AS has_featured_image_block
```
felixarntz, line 51: See above:
```suggestion
    TRUE AS has_featured_image_block
```

felixarntz (approved):
@mukeshpanchal27 I've applied the latest suggestions and added the latest query results to the PR description.

While it looks like only ~6% of block theme sites, we need to keep in mind that this is only considering the home pages of those sites. There's a good chance that a site doesn't use featured images on its home page, but it still might use them on individual articles, or only certain archives (if that's not the home page).

Unfortunately the HTTP Archive dataset is limited in that regard and therefore we may not be able to get more accurate data than this.

mukeshpanchal27:
Thanks for the update and for incorporating the suggestions!

You're absolutely right. Even if it's around 6% on home pages, the actual usage could be higher in other templates.

Given the limitations of the HTTP Archive dataset, I agree that this might be the best approximation we can get for now. Still, it's helpful as a directional indicator. Appreciate you digging into this!

PR #186, opened 2025-05-21
Introduce Markdown output format

westonruter:
In order to format the output of running various commands into Markdown for pasting in GitHub issues, I would often copy the CSV into a Google Sheet, and then copy from the sheet into GitHub. I'd then have to manually add the necessary column alignments and tidy things up (like remove the `style` tag that Google Sheets also outputs). So this PR introduces the `md` output format to go alongside the default `table` and `csv` formats.

```bash
npm run research -- benchmark-server-timing -u http://localhost:10008/bison/ -n 50 -o md
```

Output:

| URL                         | http://localhost:10008/bison/ |
| :-------------------------- | ----------------------------: |
| Success Rate                |                          100% |
| Response Time (median)      |                         20.12 |
| wp-before-template (median) |                          7.07 |
| wp-template (median)        |                         12.14 |
| wp-total (median)           |                         19.24 |

```bash
npm run research -- benchmark-web-vitals --file=urls.txt -n 50 -o md -c broadband
```

Output:

| URL               | http://localhost:10008/bison/?disable_script_fetchpriority_low=1 | http://localhost:10008/bison/ |
| :---------------- | ---------------------------------------------------------------: | ----------------------------: |
| Success Rate      |                                                             100% |                          100% |
| FCP (median)      |                                                            141.5 |                         138.8 |
| LCP (median)      |                                                            409.9 |                         379.2 |
| TTFB (median)     |                                                             35.9 |                          35.8 |
| LCP-TTFB (median) |                                                            373.6 |                         343.8 |

### DIFF
--- a/cli/commands/analyze-loading-optimization.mjs
+++ b/cli/commands/analyze-loading-optimization.mjs
@@ -76,7 +76,7 @@ export const options = [
 	},
 	{
 		argname: '-o, --output <output>',
-		description: 'Output format: csv, csv-oneline, table, or json', // TODO: Add ability to output CSV as single row for sake of putting in spreadsheet.
+		description: 'Output format: csv, csv-oneline, table, md, or json', // TODO: Add ability to output CSV as single row for sake of putting in spreadsheet.
 		defaults: OUTPUT_FORMAT_TABLE,
 	},
 ];
@@ -104,7 +104,7 @@ function getParamsFromOptions( opt ) {
 		! [ 'csv-oneline', 'json' ].includes( params.output )
 	) {
 		throw new Error(
-			`Invalid output ${ opt.output }. The output format provided via the --output (-o) argument must be either "table", "csv", "csv-oneline", or "json".`
+			`Invalid output ${ opt.output }. The output format provided via the --output (-o) argument must be either "table", "csv", "csv-oneline", "json", or "md".`
 		);
 	}
 
--- a/cli/commands/analyze-loading-optimization.mjs
+++ b/cli/commands/analyze-loading-optimization.mjs
@@ -76,7 +76,7 @@ export const options = [
 	},
 	{
 		argname: '-o, --output <output>',
-		description: 'Output format: csv, csv-oneline, table, md, or json', // TODO: Add ability to output CSV as single row for sake of putting in spreadsheet.
+		description: 'Output format: csv, csv-oneline, table, md, or json',
 		defaults: OUTPUT_FORMAT_TABLE,
 	},
 ];
--- a/cli/commands/benchmark-server-timing.mjs
+++ b/cli/commands/benchmark-server-timing.mjs
@@ -348,16 +348,5 @@ function outputResults( opt, results ) {
 		tableData.push( tableRow );
 	}
 
-	// Format the numbers with a consistent number of decimal points.
-	if ( opt.output === 'table' || opt.output === 'md' ) {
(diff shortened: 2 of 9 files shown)


cli/commands/benchmark-server-timing.mjs
```diff
@@ -348,5 +348,16 @@ function outputResults( opt, results ) {
...
+	// Format the numbers with a consistent number of decimal points.
+	if ( opt.output === 'table' || opt.output === 'md' ) {
+		tableData.forEach( ( row ) => {
+			for ( let i = 1; i < row.length; i++ ) {
+				if ( typeof row[ i ] === 'number' ) {
+					row[ i ] = row[ i ].toFixed( 2 );
+				}
+			}
+		} );
+	}
```
felixarntz, line 360: Not sure about this:
* Not great to format with a hard-coded precision. 2 is not always enough.
* It's a separate change from the overarching addition of Markdown this PR is doing anyway.

I think it would be better to consider this separately. Enabling formatting makes sense, but it needs to be flexible, so that no precision is "force lost".

cli/commands/benchmark-web-vitals.mjs
```diff
@@ -828,6 +828,18 @@ function outputResults( opt, results ) {
...
+	if ( opt.output === 'table' || opt.output === 'md' ) {
+		const precision = opt.showVariance ? 2 : 1;
+		tableData.forEach( ( row ) => {
+			for ( let i = 1; i < row.length; i++ ) {
+				if ( typeof row[ i ] === 'number' ) {
+					row[ i ] = row[ i ].toFixed( precision );
+				}
+			}
+		} );
+	}
```
felixarntz, line 841: See above.

felixarntz (commented):
@westonruter Mostly LGTM, just a few points of feedback.

cli/commands/benchmark-server-timing.mjs
westonruter: I'll revert this, but the reason for using 2 specifically is that `benchmark-server-timing` is already rounding to 2 points of precision:

https://github.com/GoogleChromeLabs/wpp-research/blob/661802d86a09510c5a4e54b867668935b3bd0f01/cli/commands/benchmark-server-timing.mjs#L276-L345

cli/commands/benchmark-web-vitals.mjs
westonruter: Removed in 986ef6e

cli/commands/benchmark-server-timing.mjs
westonruter: Removed in 986ef6e

cli/commands/analyze-loading-optimization.mjs
```diff
@@ -76,7 +76,7 @@ export const options = [
 	},
 	{
 		argname: '-o, --output <output>',
-		description: 'Output format: csv, csv-oneline, table, or json', // TODO: Add ability to output CSV as single row for sake of putting in spreadsheet.
+		description: 'Output format: csv, csv-oneline, table, md, or json', // TODO: Add ability to output CSV as single row for sake of putting in spreadsheet.
```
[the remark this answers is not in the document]
westonruter: I can't really remember 😅 

```suggestion
		description: 'Output format: csv, csv-oneline, table, md, or json',
```

felixarntz (approved):

PR #187, opened 2025-05-21
Add `--diff` argument to `benchmark-web-vitals` and `benchmark-server-timing`

westonruter:
This is a follow-up to https://github.com/GoogleChromeLabs/wpp-research/pull/186. 

The most common use case for benchmarking two URLs is to see the performance impact before/after. To do this, I would often copy the CSV output into a Google Sheet and then add columns to compute the difference in terms of milliseconds and percentage. Since this is such a common scenario (at least for me), this PR adds support for passing `--diff` when calling `benchmark-web-vitals` or `benchmark-server-timing`, if the `--file` arg is referring to a file that contains two URLs.

Example output:

```bash
npm run research -- benchmark-web-vitals --file="urls.txt" --number=100 --output="md" --network-conditions="broadband" --diff
```

| URL               | http://localhost:10008/bison/?disable_script_fetchpriority_low=1 | http://localhost:10008/bison/ | Diff (ms) | Diff (%) |
|:------------------|-----------------------------------------------------------------:|------------------------------:|----------:|---------:|
| Success Rate      |                                                             100% |                          100% |        -- |       -- |
| FCP (median)      |                                                           142.55 |                         141.7 |     -0.85 |    -0.6% |
| LCP (median)      |                                                           409.35 |                        382.35 |    -27.00 |    -6.6% |
| TTFB (median)     |                                                            34.65 |                          35.3 |     +0.65 |    +1.9% |
| LCP-TTFB (median) |                                                           374.15 |                           347 |    -27.15 |    -7.3% |

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -123,6 +123,7 @@ Sends the selected number of requests with a certain concurrency to provided URL
 * `--concurrency` (`-c`): Number of requests to make at the same time.
 * `--number` (`-n`): Total number of requests to send.
 * `--file` (`-f`): File with URLs (one URL per line) to run benchmark tests for.
+* `--diff` (`-d`): Add diffs in terms of milliseconds and percentages when two URLs are provided via `--file`.
 * `--output` (`-o`): The output format: Either "table", "csv", or "md".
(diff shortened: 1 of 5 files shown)


swissspidy (approved):
I like it! 👍

PR #188, opened 2025-05-21
Catch Puppeteer errors to count toward non-success rate in benchmark-web-vitals

westonruter:
Very often when I do `benchmark-web-vitals` while emulating slow network conditions, I'll end up in a case where the results get truncated due to an exception thrown in `benchmarkURL()` which is not caught:

```
...
Benchmarking 37 / 100...Success.
Benchmarking 38 / 100...Success.
Benchmarking 39 / 100...Success.
Benchmarking 40 / 100...Error: Navigation timeout of 30000 ms exceeded.
Benchmarking URL http://localhost:10008/bison/...
Priming network...
Benchmarking 1 / 100...Success.
Benchmarking 2 / 100...Success.
Benchmarking 3 / 100...Success.
...
```

In reality, such exceptions should count toward the non-success rate the same as a non-200 response. So this PR adds a `try`/`catch` block around the Puppeteer logic so that such errors are properly handled. The `completeRequests` counter is only incremented when no more calls to Puppeteer are needed.

This PR should be removed with whitespace changes omitted.

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -569,14 +569,7 @@ async function benchmarkURL( url, metricsDefinition, params, logProgress ) {
 			}
 
 			if ( response.status() !== 200 ) {
-				if ( logProgress ) {
-					log(
-						formats.error(
-							`Error: Bad response code ${ response.status() }.`
-						)
-					);
-				}
-				continue;
+				throw new Error( `Bad response code ${ response.status() }.` );
 			}
 
 			if ( groupedMetrics.webVitals ) {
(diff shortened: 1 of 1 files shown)


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -514,124 +514,141 @@ async function benchmarkURL( url, metricsDefinition, params, logProgress ) {
...
-		completeRequests++;
+			if ( response.status() !== 200 ) {
+				if ( logProgress ) {
+					log(
+						formats.error(
+							`Error: Bad response code ${ response.status() }.`
+						)
+					);
+				}
+				continue;
```
felixarntz, line 579: Since you're adding a central error logger in the `catch` now, it would be great to `throw` a new error here with the message instead of logging directly. Just to centralize / remove duplicated logging code :)

felixarntz (approved):
@westonruter Great catch! LGTM, just one small enhancement that would be great to address before merge.

cli/commands/benchmark-web-vitals.mjs
westonruter: Good idea. Done in 7fe49f02a8be9e820ee0985d25181c0875fbc06d

PR #189, opened 2025-05-23
Allow passing multiple `--url` args when benchmarking instead of needing to use `--file` to benchmark multiple URLs

westonruter:
I'm almost always benchmarking multiple URLs and it's annoying to have to create a file to add so that I can supply those multiple URLs via the `--file` arg. So this PR eliminates the need for a separate URLs file by allowing multiple URLs to be supplied by repeating the `--url` argument.

For example:

```bash
npm run research -- benchmark-web-vitals --number=10 --output="md" --network-conditions="broadband" \
  --url "http://localhost:10008/bison/?disable_print_script_modules_in_footer&disable_script_fetchpriority_low" \
  --url "http://localhost:10008/bison/?disable_print_script_modules_in_footer" \
  --url "http://localhost:10008/bison/?disable_script_fetchpriority_low" \
  --url http://localhost:10008/bison/
```

This involves the usage of the `parseArgs` parameter in the commander npm package, which this PR also updates to v14 from v12. (667d34514c9f74c41af2e373ca641c8e4a209f7b)

Additional changes in this PR:

1. When all benchmark requests fail for a given URL, show an error progress message rather than a success one. (3a57abcc81b7a8303cc5c4e18c02b0d6bf1cbc9d)
2. When a network priming request fails, log an error and ensure the browser instance is closed to prevent the node process from continuing indefinitely. (90486f5224a341d4809c629315c8370cf0f968cd)
3. Prevent logging `...` immediately after logging the URL so that it isn't linkified in the terminal. (462f7a94ae1585bf362b1f3ae317f07e087b78cc)

### DIFF
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -416,9 +416,9 @@ export async function handler( opt ) {
 		if ( logURLProgress ) {
 			// If also logging individual iterations, put those on a new line.
 			if ( logIterationsProgress ) {
-				log( `Benchmarking URL ${ url }...` );
+				log( `Benchmarking URL ${ url } ... ` );
 			} else {
-				logPartial( `Benchmarking URL ${ url }...` );
+				logPartial( `Benchmarking URL ${ url } ... ` );
 			}
 		}
 
--- a/cli/commands/benchmark-web-vitals.mjs
+++ b/cli/commands/benchmark-web-vitals.mjs
@@ -444,12 +444,10 @@ export async function handler( opt ) {
 					} else {
 						log( formats.success( message ) );
 					}
+				} else if ( 0 === completeRequests ) {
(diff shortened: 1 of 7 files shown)


cli/commands/benchmark-web-vitals.mjs
```diff
@@ -434,13 +438,18 @@ export async function handler( opt ) {
...
-					);
+					const message = `Completed benchmarking URL ${ url }.`;
+					if ( 0 === completeRequests ) {
+						log( formats.error( message ) );
+					} else {
+						log( formats.success( message ) );
+					}
 				} else {
-					log( formats.success( 'Success.' ) );
+					if ( 0 === completeRequests ) {
```
swissspidy, line 448: Can do an `else if` here

swissspidy (approved):
Looks good, just some lint errors

PR #194, opened 2025-07-03
Implement HTTP Archive query to assess usage of `Allow to wrap to multiple lines option` in Row block in block themes

### LINKED ISSUE #193: Query to assess usage of `Allow to wrap to multiple lines` option in Row block in block themes
As part of the [Improving the calculation of image sizes attributes project](https://github.com/WordPress/performance/issues/760), it would be helpful to understand how widely the `Allow to wrap to multiple lines` option in the **Row block** is used across WordPress sites using block themes.

This option adds the `is-nowrap` class to Row block markup, so we can detect its usage in the wild by looking for:

```html
<div class="wp-block-group is-horizontal is-nowrap
```

mukeshpanchal27:
Fixes #193

## Query results

client | urls | block_theme | total | pct_block_theme | pct_total
-- | -- | -- | -- | -- | --
desktop | 4290 | 101532 | 4382657 | 0.042252688807469567 | 0.00097885825881423067
mobile | 4980 | 131204 | 5444645 | 0.0379561598731746 | 0.00091466018445647058

Based on June 2025 dataset


sql/2025/07/row-block-allow-to-wrap-to-multiple-lines-usage.sql
```diff
@@ -0,0 +1,77 @@
...
+    client,
+    page,
+    TRUE AS has_row_block_use_wrap_option
+  FROM
+    `httparchive.crawl.requests`
+  WHERE
+    date = DATE_TO_QUERY
+    AND is_root_page = TRUE
+    AND is_main_document
+    AND REGEXP_CONTAINS(response_body, r'<div class="wp-block-group is-horizontal is-nowrap')
```
swissspidy: The group block's tag name can be dynamic, it doesn't have to be a `<div>`. See https://github.com/WordPress/gutenberg/blob/8a5a63a4f284e11776a8d3881ec9b14a8965cb18/packages/block-library/src/group/edit.js#L35-L54

Also, is the list class names always deterministic like that? What if they're in a different order?

sql/2025/07/row-block-allow-to-wrap-to-multiple-lines-usage.sql
```diff
@@ -0,0 +1,77 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# See query results here: https://github.com/GoogleChromeLabs/wpp-research/pull/194
+
+DECLARE
+  DATE_TO_QUERY DATE DEFAULT '2025-07-01';
```
swissspidy: I don't think this one is available yet, but we can test with the previous one for now.

sql/2025/07/row-block-allow-to-wrap-to-multiple-lines-usage.sql
mukeshpanchal27: Thanks for pointing it out i missed that.

Query updated.

swissspidy:
Haven't really reviewed the query yet, but here are the results:

client | urls | block_theme | total | pct_block_theme | pct_total
-- | -- | -- | -- | -- | --
desktop | 4290 | 101532 | 4382657 | 0.042252688807469567 | 0.00097885825881423067
mobile | 4980 | 131204 | 5444645 | 0.0379561598731746 | 0.00091466018445647058

felixarntz (approved):
Thanks @mukeshpanchal27, the query looks solid to me. 👍

